PR #1, opened 2017-07-19
DM-11118: Build stubbed out verify_ap

kfindeisen:
This is an initial implementation for `verify_ap`. It should be equivalent to the pseudocode on [DM-11118](https://jira.lsstcorp.org/browse/DM-11118), except that it has two afterburners and adds explicit support for `lsst.verify`.

The current code does not run because there are no working datasets, so the constructor of `Dataset` always raises. In addition, the stubs in `pipedriver` all raise `NotImplementedError`. However, `test_args` does run, and the main program's command-line parser is functional:

    $ python python/lsst/verify/ap/verify_ap.py -h
    usage: verify_ap.py [-h] --dataIdString DATAID [-j PROCESSES] [--silent]
                        --dataset {HiTS} (--output OUTPUT | --rerun OUTPUT)
                        [--version]

    Executes the LSST DM AP pipeline and analyzes its performance using metrics.

    optional arguments:
      -h, --help            show this help message and exit
      --dataIdString DATAID
                            An identifier for the data to process. Syntax and
                            meaning same as --id in Tasks.
      -j PROCESSES, --processes PROCESSES
                            Number of processes to use
      --silent              Do NOT submit metrics to SQuaSH (not yet implemented).
      --dataset {HiTS}      The source of data to pass through the pipeline.
      --output OUTPUT       The location of the repository to use for program
                            output.
      --rerun OUTPUT        The location of the repository to use for program
                            output, as DATASET/rerun/OUTPUT
      --version             show program's version number and exit

### DIFF
--- a/tests/test_args.py
+++ b/tests/test_args.py
@@ -1,0 +1,108 @@
+#
+# LSST Data Management System
+# Copyright 2017 LSST Corporation.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+from __future__ import absolute_import, division, print_function
+
+import shlex
+import unittest
+
+import lsst.utils.tests
+import lsst.ap.verify.ap_verify as ap_verify
+
+
+class CommandLineTestSuite(lsst.utils.tests.TestCase):
+
+    def _parseString(self, command_line):
+        """Tokenize and parse a command line string.
+
+        Parameters
+        ----------
+        command_line: `str`
+            a string containing Unix-style command line arguments, but not the
+            name of the program
+        """
+        return ap_verify._VerifyApParser().parse_args(shlex.split(command_line))
+
+    def test_missing(self):
+        """Verify that a command line consisting missing required arguments is rejected.
+        """
+        args = '--dataset HiTS2015 --output tests/output/foo'
+        with self.assertRaises(SystemExit):
+            self._parseString(args)
+
+    def test_minimum(self):
+        """Verify that a command line consisting only of required arguments parses correctly.
+        """
+        args = '--dataset HiTS2015 --output tests/output/foo --dataIdString "visit=54123"'
+        parsed = self._parseString(args)
+        self.assertIn('dataset', dir(parsed))
+        self.assertIn('output', dir(parsed))
+        self.assertIn('dataId', dir(parsed))
+
+    def test_rerun(self):
+        """Verify that a command line with reruns is handled correctly.
+        """
+        args = '--dataset HiTS2015 --rerun me --dataIdString "visit=54123"'
+        parsed = self._parseString(args)
+        out = ap_verify._get_output_dir('non_lsst_repo/', parsed.output, parsed.rerun)
+        self.assertEqual(out, 'non_lsst_repo/rerun/me')
+
+    def test_rerun_input(self):
+        """Verify that a command line trying to redirect input is rejected.
+        """
+        args = '--dataset HiTS2015 --rerun from:to --dataIdString "visit=54123"'
+        with self.assertRaises(SystemExit):
+            self._parseString(args)
+
+    def test_two_outputs(self):
+        """Verify that a command line with both --output and --rerun is rejected.
+        """
+        args = '--dataset HiTS2015 --output tests/output/foo --rerun me --dataIdString "visit=54123"'
+        with self.assertRaises(SystemExit):
+            self._parseString(args)
+
+    def test_bad_dataset(self):
+        """Verify that a command line with an unregistered dataset is rejected.
+        """
+        args = '--dataset FooScope --output tests/output/foo --dataIdString "visit=54123"'
+        with self.assertRaises(SystemExit):
+            self._parseString(args)
+
+    def test_bad_key(self):
+        """Verify that a command line with unsupported arguments is rejected.
+        """
+        args = '--dataset HiTS2015 --output tests/output/foo --dataIdString "visit=54123" --clobber'
+        with self.assertRaises(SystemExit):
+            self._parseString(args)
+
+
+class MemoryTester(lsst.utils.tests.MemoryTestCase):
+    pass
+
+
+def setup_module(module):
+    lsst.utils.tests.init()
+
+
+if __name__ == "__main__":
+    lsst.utils.tests.init()
+    unittest.main()
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -1,0 +1,109 @@
+#
+# LSST Data Management System
+# Copyright 2017 LSST Corporation.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+from __future__ import absolute_import, division, print_function
+
+import os
+import shutil
+import unittest
+
+import lsst.utils.tests
+from lsst.ap.verify.dataset import Dataset
+
+
+class DatasetTestSuite(lsst.utils.tests.TestCase):
+
+    def setUp(self):
+        self._testbed = Dataset('HiTS2015')
+
+    def test_init(self):
+        """Verify that if a Dataset object exists, the corresponding data are available.
+        """
+        # EUPS does not provide many guarantees about what setting up a package means
+        self.assertIsNotNone(os.getenv('AP_VERIFY_HITS2015_DIR'))
+
+    def test_datasets(self):
+        """Verify that a Dataset knows its supported datasets.
+        """
+        datasets = Dataset.get_supported_datasets()
+        self.assertIn('HiTS2015', datasets)  # assumed by other tests
+
+        # Initializing another Dataset has side effects, alas, but should not
+        # invalidate tests of whether HiTS2015 has been loaded
+        for dataset in datasets:
+            Dataset(dataset)
+
+    def test_directories(self):
+        """Verify that a Dataset reports the desired directory structure.
+        """
+        root = self._testbed.dataset_root
+        self.assertEqual(self._testbed.data_location, os.path.join(root, 'raw'))
+        self.assertEqual(self._testbed.calib_location, os.path.join(root, 'calib'))
+        self.assertEqual(self._testbed.template_location, os.path.join(root, 'templates'))
+
+    def test_output(self):
+        """Verify that a Dataset can create an output repository as desired.
+        """
+        test_dir = os.path.dirname(__file__)
+        output = os.path.join(test_dir, 'hitsOut')
+        self.assertFalse(os.path.exists(output), 'Output test invalid if directory exists.')
+
+        try:
+            self._testbed.make_output_repo(output)
+            self.assertTrue(os.path.exists(output), 'Output directory must exist.')
+            self.assertTrue(os.listdir(output), 'Output directory must not be empty.')
+            self.assertTrue(os.path.exists(os.path.join(output, '_mapper')),
+                            'Output directory must have a _mapper file.')
+        finally:
+            if os.path.exists(output):
+                shutil.rmtree(output)
+
+    def test_bad_output(self):
+        """Verify that a Dataset will not create an output directory if it is unsafe to do so.
+        """
+        test_dir = os.path.dirname(__file__)
+        output_dir = os.path.join(test_dir, 'badOut')
+
+        try:
+            os.makedirs(output_dir)
+            output = os.path.join(output_dir, 'foo.txt')
+            with open(output, 'w') as dummy:
+                dummy.write('This is a test!')
+
+            with self.assertRaises(IOError):
+                self._testbed.make_output_repo(output_dir)
+        finally:
+            if os.path.exists(output_dir):
+                shutil.rmtree(output_dir)
+
+
+class MemoryTester(lsst.utils.tests.MemoryTestCase):
+    pass
+
+
+def setup_module(module):
+    lsst.utils.tests.init()
+
+
+if __name__ == "__main__":
+    lsst.utils.tests.init()
+    unittest.main()
--- a/README.md
+++ b/README.md
@@ -3,3 +3,33 @@
 This package manages end-to-end testing and metric generation for the LSST DM Alert Production pipeline. Metrics are tested against both project- and lower-level requirements, and will be deliverable to the SQuaSH metrics service.
 
 `ap_verify` is part of the LSST Science Pipelines. You can learn how to install the Pipelines at https://pipelines.lsst.io/install/index.html.
+
+## Configuration
+
+`ap_verify` is configured from `config/dataset_config.yaml`. The file currently must have a single dictionary named `datasets`, which maps from user-visible dataset names to the eups package that implements them (see `Setting Up a Package`, below). Other configuration options may be added in the future.
+
+### Setting Up a Package
+
+`ap_verify` requires that all data be in a [dataset package](https://github.com/lsst-dm/ap_verify_dataset_template). It will create a workspace modeled after the package's `data` directory, then process any data found in the `raw` and `ref_cats` in the new workspace. Anything placed in `data` will be copied to a `ap_verify` run's workspace as-is, and must at least include a `_mapper` file naming the CameraMapper for the data.
+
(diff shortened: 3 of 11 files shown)


python/lsst/verify/ap/verify_ap.py
```diff
@@ -0,0 +1,136 @@
...
+import os
+import re
+
+import lsst.log
+from lsst.verify.ap.dataset import Dataset
+from lsst.verify.ap.metrics import MetricsParser, check_squash_ready, AutoJob
+from lsst.verify.ap.pipedriver import ApPipeParser, run_ap_pipe
+
+
+class _VerifyApParser(argparse.ArgumentParser):
```
djreiss: I'm not yet up to speed on the verify framework or the verify_ap design discussions, but is there any reason that this stuff is not set up as a `lsst.pipe.base.Task` subclass? That way you'd already have argument and config file parsing built-in.

tests/test_args.py
```diff
@@ -0,0 +1,69 @@
...
+import shlex
+import unittest
+
+import lsst.utils.tests
+import lsst.verify.ap.verify_ap as verify_ap
+
+
+class CommandLineTestSuite(lsst.utils.tests.TestCase):
+
+    def test_minimum(self):
```
djreiss: Please add docstrings to this and all `test_*` methods.

python/lsst/verify/ap/metrics.py
```diff
@@ -0,0 +1,157 @@
...
+        """Packages all metric measurements performed during this run.
+
+        The measurements shall be exported to `verify_ap.verify.json`, and the
+        metrics framework shall be shut down. If the context was exited
+        normally and the appropriate flag was passed to this object's
+        constructor, the measurements shall be sent to SQuaSH.
+        """
+        log = lsst.log.Log.getLogger('verify.ap.metrics.AutoJob.__exit__')
+
+        out_file = 'verify_ap.verify.json'
```
kfindeisen: It's the default name assumed by `verify`, according to SQR-019, and there doesn't seem an obvious reason to let the user create multiple dump files.

python/lsst/verify/ap/pipedriver.py
```diff
@@ -0,0 +1,193 @@
...
+
+__all__ = ["ApPipeParser", "run_ap_pipe"]
+
+import argparse
+
+import lsst.log
+from lsst.verify.ap.dataset import Dataset
+
+
+def _ingest_raws(dataset, repo_root):
```
kfindeisen: I couldn't think of any particular program state that would be guaranteed by having a `PipeWrapper` object sitting around, like there is with `Dataset` and `AutoJob`.

I guess I could create a class as a convenience wrapper for the function arguments, but that seems a bit gratuitous. It would also make it more difficult for other coders to implement the stubs and/or keep straight which information is needed for which pipeline step.

python/lsst/verify/ap/verify_ap.py
kfindeisen: Simon did not want `verify_ap` to be a Task, because some of the processing steps are not Tasks and so config info would be lost anyway (or at least, that's what my notes say; is that too garbled, @SimonKrughoff?)

Also, the desired behavior for `verify_ap`, particularly the idea of a "dataset" that is *not* necessarily a Butler repository (to support native observatory output, IIRC), leads to command-line arguments that are incompatible with Task requirements. This includes using `--dataIdString` instead of `--id`, and the absence of an input-repo or `--rerun INPUT:OUTPUT` argument. I'll admit I don't know the reason for avoiding `--id`, but there was some discussion about input reruns on Slack.

python/lsst/verify/ap/pipedriver.py
djreiss: To me it seems a natural case for a class, where future implementations can override the stubs in a subclass. I know you are not a fan of overuse of inheritance, but to me this seems like a situation where it would be useful.

djreiss: Oh, and I think this module could use some module-level docstring explaining the usage and where/how it gets called. (or a class-level docstring if we go with my previous suggestion).

kfindeisen: I don't think there's any plan for "future implementations"; the stubs will be filled in once `ap_pipe`'s API is finalized. Where would the subclasses live, anyway? There shouldn't be any packages that depend on `verify_ap` because it's a command-line program rather than a library...

kfindeisen: After discussing with @djreiss, this module will be redesigned to make it easier to adapt to other pipelines if somebody decides our system is better than `ci_hsc` or `validate_drp`.

tests/test_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+class DatasetTestSuite(lsst.utils.tests.TestCase):
+
+    def setUp(self):
+        self._testbed = Dataset('HiTS2015')
+
+    def test_init(self):
+        """Verify that if a Dataset object exists, the corresponding data are available.
+        """
+        try:
+            import lsst.verify.ap.hits2015  # noqa: F401
```
kfindeisen: @djreiss, I don't think this will work with what you did for DM-11117. Do you have any suggestions for how to make sure a dataset package has been set up correctly?

python/lsst/verify/ap/appipe.py
```diff
@@ -0,0 +1,159 @@
...
+    """An argument parser for data needed by ap_pipe activities.
+
+    This parser is not complete, and is designed to be passed to another parser
+    using the `parent` parameter.
+    """
+
+    def __init__(self):
+        # Help and documentation will be handled by main program's parser
+        argparse.ArgumentParser.__init__(self, add_help=False)
+        self.add_argument('--dataIdString', dest='dataId', required=True,
```
mrawls: What is the rationale for calling this `dataIdString` and not just `dataId` or `id`? The README example has it set to "visit=54123" and I'm wondering what other information might typically be passed here aside from a visit number.

python/lsst/verify/ap/pipeline.py
```diff
@@ -0,0 +1,74 @@
...
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+from __future__ import absolute_import, division, print_function
+
+
+class Pipeline(object):
```
mrawls: I'm a bit confused about the point of the Pipeline class. It's here for future-proofing but not presently used? Is that the idea?

python/lsst/verify/ap/appipe.py
```diff
@@ -0,0 +1,159 @@
...
+    ----------
+    dataset: `dataset.Dataset`
+        The dataset on which the pipeline will be run.
+    working_dir: `str`
+        The repository in which temporary products will be created. Must be
+        compatible with `dataset`.
+    parsed_cmd_line: `argparse.Namespace`
+        Command-line arguments, including all arguments supported by `ApPipeParser`.
+    """
+
```
mrawls: Just to confirm, your plan is to have each of the methods below (`_ingest_raws`, etc.) call the appropriate piece of `ap_pipe` and get ride of the `NotImplementedError`s, yes? I like how you've broken them out into logical steps and I like the names you chose 👍

python/lsst/verify/ap/appipe.py
```diff
@@ -0,0 +1,159 @@
...
+    def _ingest_calibs(self):
+        """Ingest the raw calibrations for use by LSST.
+
+        The original calibration directory shall not be modified.
+        """
+        # use self.dataset and self.repo
+        raise NotImplementedError
+
+    def _ingest_templates(self):
+        """Ingest precomputed templates for use by LSST.
```
mrawls: You might want to specify that the template can be either a calexp or a coadd (there may be more official names for these two data products, but `ap_pipe` will accept either, with a coadd being the default)

python/lsst/verify/ap/appipe.py
```diff
@@ -0,0 +1,159 @@
...
+        self.add_argument('--dataIdString', dest='dataId', required=True,
+                          help='An identifier for the data to process. '
+                          'Syntax and meaning same as --id in Tasks.')
+        self.add_argument("-j", "--processes", default=1, type=int, help="Number of processes to use")
+
+
+class ApPipe(Pipeline):
+    """Wrapper for `lsst.ap.pipe` that executes all steps through source
+    association.
+
```
mrawls: It might be worthwhile to mention that the steps must be executed in the order presented for the pipeline to work. This is pretty clear from the `run` method but it can't hurt to say upfront.

python/lsst/verify/ap/dataset.py
```diff
@@ -0,0 +1,218 @@
...
+
+                Dataset._dataset_config = dataset_map
+            except (KeyError, TypeError, yaml.YAMLError) as e:
+                raise_from(RuntimeError('Invalid config file.'), e)
+
+        return Dataset._dataset_config
+
+    @property
+    def dataset_root(self):
+        """The parent directory containing everything related to the dataset.
```
mrawls: Are we not assuming that the `data_location`, `calib_location`, and `template_location` must all be in some default set of locations that can be known just by providing `dataset_root`?

python/lsst/verify/ap/dataset.py
```diff
@@ -0,0 +1,218 @@
...
+        if not os.path.exists(self.template_location):
+            raise RuntimeError('Dataset is missing template directory at ' + self.template_location)
+        if not os.path.exists(self._stub_input_repo):
+            raise RuntimeError('Dataset at ' + self.dataset_root + 'is missing stub repo')
+        if not os.path.exists(os.path.join(self._stub_input_repo, '_mapper')):
+            raise RuntimeError('Stub repo at ' + self._stub_input_repo + 'is missing mapper file')
+
+    def make_output_repo(self, output_dir):
+        """Set up a directory as an output repository compatible with this dataset.
+
```
mrawls: Does this play nice with reruns? (Asking for a friend who really likes reruns)

python/lsst/verify/ap/verify_ap.py
```diff
@@ -0,0 +1,156 @@
...
+            add_help=True)
+        self.add_argument('--dataset', choices=Dataset.get_supported_datasets(), required=True,
+                          help='The source of data to pass through the pipeline.')
+
+        output = self.add_mutually_exclusive_group(required=True)
+        output.add_argument('--output', help='The location of the repository to use for program output.')
+        output.add_argument(
+            '--rerun', metavar='OUTPUT',
+            type=_FormattedType('[^:]+',
+                                'Invalid name "%s"; verify_ap does not support reruns as input.'),
```
mrawls: This is confusing because it's an output argument but the message is "verify_ap does not support reruns as **input**". Confused.

python/lsst/verify/ap/appipe.py
kfindeisen: The issue description said:
> "--dataIdString" (note: often just "--Id" as a command line string in other places)

which I took to mean "we have a specific reason for calling it `--dataIdString` and especially not `--id`". Do you remember where that came from?

python/lsst/verify/ap/appipe.py
kfindeisen: I'm not sure I understand. The steps *are* executed in the order determined by `run`; who (except somebody modifying that method) would need to care about what happens if you change the order? From the caller's point of view they just call `run`, and the pipeline runs.

python/lsst/verify/ap/appipe.py
kfindeisen: Yes, that's exactly the plan.

python/lsst/verify/ap/appipe.py
kfindeisen: Wait, I'm confused. Isn't the stuff that's being ingested *not* in a native LSST format? Isn't that the point of ingesting?

python/lsst/verify/ap/dataset.py
kfindeisen: Reruns are translated to directories in the `verify_ap` module, so this class doesn't need to care.

python/lsst/verify/ap/dataset.py
kfindeisen: Yes, we are...? I don't think I understand the question. :confused:

python/lsst/verify/ap/verify_ap.py
kfindeisen: The error message is for people who ignore/forget the fact that it's an output argument. Would it be more readable if the `_FormattedType` object were factored into a separate variable?

python/lsst/verify/ap/pipeline.py
kfindeisen: Sort of. It's for clarifying exactly what an alternative to `ApPipe` needs to look like (i.e., it must work in terms of a `Dataset` and an output repo, and it must have a `run` method), and dividing things needed by *any* pipeline wrapper from things specific to wrapping `ap_pipe`.

It's a very C++ic way of approaching the problem, but @parejkoj assures me it's not discouraged in Python (though he did suggest some changes if I keep this class).

python/lsst/verify/ap/appipe.py
kfindeisen: (Please disregard my previous comment here.) Am I correct in assuming that raws and calibs should be in a telescope-specific format rather than LSST types like "calexp"?

python/lsst/verify/ap/appipe.py
mrawls: Hmm, the distinction is likely because it's not a canonical `dataId`, and `id` alone is a vague term. I'm fine with `dataIdString`, I just mistakenly thought the name came from you and not from the design outline. No problem.

kfindeisen: How is it not a canonical `dataId`? I wrote the documentation assuming it was (i.e., `'Syntax and meaning same as --id in Tasks.'`).

python/lsst/verify/ap/appipe.py
mrawls: That's true; I was thinking of a very weird case where somebody stumbled across `ApPipe`, ignored the run method, and thought "I'll just run `process`" and ignore all the rest!" It's fine as-is.

python/lsst/verify/ap/appipe.py
mrawls: It should probably be whatever the butler calls them, which is telescope-specific for DECam CP calibration products (e.g., `cpBias` and `cpFlat`) and shouldn't be telescope-specific for the other products (`raw`, `calexp`, etc.). I think the correct template datatypes are either `calexp` or `deepCoadd_psfMatchedWarp`.

python/lsst/verify/ap/appipe.py
mrawls: My understanding is that `dataId` means a very specific thing to the butler and it's represented as a dictionary in python. In contrast, this is a string with some of the info you might find in a "proper" `dataId`.

kfindeisen: So what should the user-facing documentation say? "May not support all features of a Butler dataId, exact subset TBD"?

python/lsst/verify/ap/dataset.py
mrawls: I think the code doesn't make it immediately clear that `dataset_root` will know where to find "lower-level" things like `data_location`, `calib_location`, etc. But I guess there's probably not another obvious way to define them all, so it's fine.

python/lsst/verify/ap/verify_ap.py
mrawls: I think I'm still confused as to what the `type` argument is catching here. Normally you use it to clarify that the argparse input must be an int, str, float, whatever. But you're trying to catch some kind of user error where they mistakenly think `--rerun` could be used in lieu of `--dataset`?

python/lsst/verify/ap/appipe.py
mrawls: Yeah, something like that would be great :) but at least give a list of what it currently does support (visit, obviously; anything else?)

python/lsst/verify/ap/appipe.py
mrawls: My answer was template-specific but your question was about raws and calibs, sorry. I'm actually not sure, this would be a good question for slack.

python/lsst/verify/ap/verify_ap.py
kfindeisen: The [argparse documentation](https://docs.python.org/2/library/argparse.html#type) makes it sound like `type` is where input validation is supposed to go, given that they imply that it would be the appropriate place for range-checking if `choices` didn't exist. I agree it feels weird, so maybe I misunderstood.

The error I'm trying to catch is a user saying `--rerun INPUT:OUTPUT`, which is common(?) when using Tasks. I'm guessing the mindset behind the error would be less "in lieu of `--dataset`" and more "`--dataset` specifies the survey but not the file location".

python/lsst/verify/ap/appipe.py
kfindeisen: What's supported is up to you. From the design outline:
> ap_pipe.process(repo_location, rerun, parallelization_level, dataIdString)
> ap_pipe.diff(repo_location, rerun, parallelization_level, dataIdString)

mrawls: Fair enough! For now it will just be visit, but I see why this is more of an `ap_pipe` consideration. Let's just leave this be.

python/lsst/verify/ap/verify_ap.py
mrawls: Oh, OK then! I'd never seen anyone try `--rerun INPUT:OUTPUT` but now that you say that and I pretend to remember how regexes work I can see what you've done. Thanks for clarifying. So you're really insisting that there not be more than one thing shoved into `--rerun`. A clearer error message might be `Invalid name "%s"; verify_ap supports only one output destination when using --rerun. You have entered something that appears to be of the form INPUT:OUTPUT. Try again and specify only OUTPUT.`

python/lsst/verify/ap/dataset.py
kfindeisen: I'm not sure if this answers your question, but I don't expect `ApPipe` will need to use `dataset_root` at all. That's just there as a backup for when an operation really needs to know about the dataset as a whole (currently, it's only used to figure out reruns).

python/lsst/verify/ap/pipeline.py
djreiss: The Pipeline class was my suggestion -- and thanks for doing it. I also think all the "Use ..." comments in each method is helpful, which I failed to realize would disappear when you put the methods into a class.

tests/test_dataset.py
djreiss: Apparently I need to add it to repos.yaml. I'll do that shortly. Until I do, You can also just clone the dataset repo and use `setup -r .`. Let me know if that does/doesn't work.

PR #3, opened 2017-08-10
DM-10779: Implement running time metric(s)

kfindeisen:
This PR adds a function for extracting running time from Task metadata, and adds some infrastructure to `ap_verify` to support it and future features.


python/lsst/ap/verify/measurements/api.py
```diff
@@ -0,0 +1,64 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+"""Code for measuring software performance metrics.
```
ebellm: Could this be called something more descriptive than "api.py?"

python/lsst/ap/verify/measurements/performance.py
```diff
@@ -0,0 +1,71 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+__all__ = ["measure_runtime"]
+
+import astropy.units as u
+
+import lsst.verify
+
+
```
ebellm: similarly, could this be called something more descriptive than "performance.py?"

kfindeisen:
I am willing to replace the `ApPipe` class with the function-based module that came before it -- the former was a request by @djreiss in the previous review that, as Lupton pointed out yesterday, constitutes scope creep. Note that the `ap_pipe` wrapper is outside the scope of this ticket, so it may be easier (particularly for @mrawls) if we create a new ticket to make that change.

I must strongly object to dumping the `ApPipe` code into `run_ap_verify`, however. This would not only bloat that function well beyond its already long length and make the code more complicated, but add extra, implementation-specific dependencies between the main `ap_verify` module and `ap_pipe`. If we want any hope of having `ap_verify` be easy to maintain and improve in the future, everything that depends on `ap_pipe` needs to be in its own module, safely compartmentalized from the rest of the system rather than mixed in with code related to argument and metrics management

python/lsst/ap/verify/measurements/performance.py
kfindeisen: Such as? `measurements/performance.py` seems to quickly summarize what the code is for.

python/lsst/ap/verify/measurements/api.py
kfindeisen: I couldn't think of one, and still can't. This file's purpose is to provide a single point of entry to the `measurements` sub-package, so that `ap_verify.py` doesn't need to know about implementation details like "which measurements have actually been implemented"?

Is there a name that communicates that to you?

ebellm: What about `compute_metrics`?  "API" suggests to me a library with multiple methods, not an abstraction layer.

A top-level docstring which better indicated the purpose of the package would help here--the code for measuring metrics is not actually in this file.

python/lsst/ap/verify/measurements/performance.py
ebellm: It depends where you plan to put other metrics code--it's all "performance" at some level.  If you're thinking of small packages, `runtime_metrics.py` would be more clear.

ebellm:
Okay, I'll let you refactor the code to be more functional, but I don't agree that `ap_verify` needs to be strictly compartmentalized from the details of `ap_pipe`--it should just be a small wrapper to add the metrics and verification support.  I'd prefer tighter coupling to keep the package lightweight.

python/lsst/ap/verify/measurements/performance.py
kfindeisen: "runtime" to me sounds like "not compile-time", but I see your point. How about `code_profiling.py`?

(I'd prefer to avoid 'metrics' or 'measurement' in the name to avoid Department of Redundancy Department).

ebellm: okay, that works for me.

PR #4, opened 2017-08-29
DM-11575: Update packaging layout for ap_verify

djreiss:


python/lsst/ap/verify/dataset.py
```diff
@@ -171,7 +171,7 @@ def refcat_location(self):
         -------
         a string giving the location of the top-level directory for astrometric and photometric catalogs
         """
-        return os.path.join(self.dataset_root, 'ref_cats')
+        return os.path.join(self.dataset_root, 'refcats')
```
kfindeisen: Please update `test_dataset.py` so that it passes. Please check that the unit tests do pass (requires `eups declare` or `eups setup` on `ap_verify_hits2015`).

PR #12, opened 2017-11-29
DM-11592: Better documentation for ap_verify

kfindeisen:
This PR adds new-style documentation to the `ap_verify` package. Since the package is not a library and has a minimal public-facing API, most of the documentation is instructions for `ap_verify` users and guides to key `ap_verify` concepts (in particular, datasets).

There are a number of cross-references to the other `ap` packages, the `verify` documentation, the `pipe_base` documentation, and the butler and obs framework topic documentation. Naturally, these references will need to be updated once these documents actually exist.

Since new-style documentation is not yet built by Travis, I don't have an HTML version of the documentation to link to.


doc/ap_verify/index.rst
```diff
@@ -0,0 +1,27 @@
+.. _ap_verify-package:
+
+#########
+ap_verify
+#########
+
+The ``ap_verify`` package wraps :py:mod:`ap_pipe <lsst.ap.pipe>` with support for managing :py:mod:`verify <lsst.verify>` metrics.
```
jonathansick: For "verify" use the actual module namespace for clarity:

```rst
for managing `lsst.verify` metrics.
```

Note that single back-ticks can be used for any type of Python API reference (being the default role).

doc/lsst.ap.verify/datasets-install.rst
```diff
@@ -0,0 +1,41 @@
...
+###################
+
+:ref:`ap-verify-datasets` packages data in self-contained units that are intended to be easy to install for LSST Stack users.
+It is not necessary to install all datasets supported by ``ap_verify``, only those you intend to use.
+
+Prerequisites
+-------------
+
+The Dataset framework requires that the computer have version 13.0 or later of the LSST Stack (specifically, the ``obs`` packages and their dependencies) installed.
+`Installing lsst_distrib <https://pipelines.lsst.io/#installation>`_ is the simplest way to ensure all dependencies are satisfied.
```
jonathansick:
```rst
:ref:`Installing lsst_distrib <part-installation>` [...]
```

doc/lsst.ap.verify/datasets-install.rst
```diff
@@ -0,0 +1,41 @@
...
+
+The framework also requires `Git LFS`_ and the `EUPS`_ package management system. The latter is usually included by the Stack installation process.
```
jonathansick: Second sentence:

```
EUPS is included in the Stack installation.
```

doc/lsst.ap.verify/datasets-install.rst
```diff
@@ -0,0 +1,41 @@
...
+
+Installation Procedure
+----------------------
+
+Use Git LFS to clone the desired dataset's GitHub repository.
+To get the URL, see the :ref:`package documentation<ap-verify-datasets-index>` for the dataset in question.
+
+.. TODO: should we have a proper versioning system for datasets? (DM-12853)
+
+Once the dataset has been installed, use ``eups declare`` to register the downloaded directory.
```
jonathansick: Try to wrap inline shell commands in the `command` role, like

```rst
:command:`eups declare`
```

doc/lsst.ap.verify/datasets-install.rst
```diff
@@ -0,0 +1,41 @@
...
+The product name given to EUPS must match the repository name; the version can be anything.
+It is also possible to register the dataset using ``setup``, but this is recommended only for temporary tests.
+
+For example, to install the :ref:`HiTS 2015 <ap_verify_hits2015-package>` dataset,
+
+.. code-block:: sh
+
+   $ git clone https://github.com/lsst/ap_verify_hits2015 mydata
```
jonathansick: I've been leaving out any `$` or `>` characters to reduce copy-paste errors. There are special directives for making uncopyable prompts but I've had some trouble with them.

doc/lsst.ap.verify/running.rst
```diff
@@ -0,0 +1,80 @@
...
+This page describes the minimum options needed to run ``ap_verify``.
+For more details, see the :ref:`ap-verify-cmd` or run ``ap_verify`` with the ``-h`` argument.
+
+.. _ap-verify-dataset-name:
+
+Datasets as Input Arguments
+---------------------------
+
+While ``ap_verify`` is not a :ref:`command-line task<lsst.pipe.base>`, the command-line interface is designed to resemble that of command-line tasks where practical.
+The most important difference is that, since ``ap_verify`` begins with an uningested :ref:`dataset<ap-verify-datasets>`, the input argument is a dataset name rather than a repository.
```
jonathansick:
```
Since ``ap_verify`` begins with an uningested :ref:`dataset<ap-verify-datasets>`, the input argument is a dataset name rather than a repository.
```

doc/lsst.ap.verify/running.rst
```diff
@@ -0,0 +1,80 @@
...
+``ap_verify`` is a Python script designed to be run on both developer machines and verification servers.
+This page describes the minimum options needed to run ``ap_verify``.
+For more details, see the :ref:`ap-verify-cmd` or run ``ap_verify`` with the ``-h`` argument.
+
+.. _ap-verify-dataset-name:
+
+Datasets as Input Arguments
+---------------------------
+
+While ``ap_verify`` is not a :ref:`command-line task<lsst.pipe.base>`, the command-line interface is designed to resemble that of command-line tasks where practical.
```
jonathansick: I'd move the first sentence up to being the second sentence on this page.

doc/lsst.ap.verify/running.rst
```diff
@@ -0,0 +1,80 @@
...
+Here:
+
+* ``HiTS2015`` is the :ref:`dataset name <ap-verify-dataset-name>`,
+* ``workspace/hits/`` is the location of the :ref:`Butler repository<command-line-task-data-repo-using-uris>` in which the pipeline will work,
+* ``visit=54123 ccd=25 filter=g`` is the :ref:`dataId<command-line-task-dataid-howto-about-dataid-keys>` to process, and
+* ``--silent`` disables SQuaSH metrics reporting.
+
+This will create a workspace (a Butler repository) in ``workspace/hits`` based on ``<hits-data>/data/``, ingest the HiTS data into it, then run visit 54123 through the entire AP pipeline.
+
+.. note::
```
jonathansick: No need for this note, I think. You already said it's similar, and this page along with the command reference, documents what it is exactly.

doc/lsst.ap.verify/running.rst
```diff
@@ -0,0 +1,80 @@
...
+Since datasets are :ref:`not, in general, repositories<ap-verify-datasets-butler>`, the ``--rerun`` argument only superficially resembles the analogous argument for command-line tasks.
+In particular, ``ap_verify``'s ``--rerun`` argument does not support repository chaining (as in ``--rerun input:output``); the input for ``ap_verify`` will always be determined by the ``--dataset``.
+
+.. _ap-verify-results:
+
+How to Use Measurements of Metrics
+----------------------------------
+
+After ``ap_verify`` has run, it will produce a file named ``ap_verify.verify.json`` in the working directory.
+This file contains metric measurements in ``lsst.verify`` format, and can be loaded and read as described in the :ref:`verify documentation<lsst.verify>` or in `SQR-019 <https://sqr-019.lsst.io>`_.
```
jonathansick: ``lsst.verify documentation`` (for clarify, refer to other packages by their full namespace)

doc/lsst.ap.verify/running.rst
```diff
@@ -0,0 +1,80 @@
...
+
+Unless the ``--silent`` argument is provided, ``ap_verify`` will also upload measurements to the `SQuaSH service <https://squash.lsst.codes/>`_ on completion.
+See the SQuaSH documentation for details.
+
+If the pipeline is interrupted by a fatal error, completed measurements will be saved to ``ap_verify.verify.json`` for debugging purposes, but nothing will get sent to SQuaSH.
```
jonathansick: Might want to add the word "file" to the sentence:

```rst
save to a :file:`ap_verify.verify.json` file [...]
```

doc/lsst.ap.verify/command-line-reference.rst
```diff
@@ -0,0 +1,139 @@
+.. _ap-verify-cmd:
+
+######################
+Command Line Reference
```
jonathansick: Command-Line Reference

doc/lsst.ap.verify/command-line-reference.rst
```diff
@@ -0,0 +1,139 @@
+.. _ap-verify-cmd:
```
jonathansick: I think it'd be useful to add

```rst
.. program:: ap_verify.py
```

here.

This way the options documented below can be uniquely referenced as

```rst
:option:`ap_verify.py --dataset`
```

or

```rst
:option:`--dataset <ap_verify.py --dataset>`
```

for short

(http://www.sphinx-doc.org/en/stable/domains.html?highlight=option#directive-program)

doc/lsst.ap.verify/command-line-reference.rst
```diff
@@ -0,0 +1,139 @@
...
+
+   **Print help.**
+
+   The help is equivalent to this documentation page, describing command-line arguments.
+
+.. option:: -j <processes>, --processes <processes>
+
+   **Number of processes to use.**
+
+   When processes is larger than 1 the pipeline may use the Python `multiprocessing` module to parallelize processing of multiple datasets across multiple processors.
```
jonathansick: ``1``

doc/lsst.ap.verify/command-line-reference.rst
```diff
@@ -0,0 +1,139 @@
+.. _ap-verify-cmd:
+
+######################
+Command Line Reference
+######################
+
+This page describes the command-line arguments and environment variables used by ``ap_verify``.
```
jonathansick: You might want to distinguish between `ap_verify` the EUPS package and `ap_verify` the command-line executable.

One way might be to have the EUPS package reference link to the package homepage.

Then the command-line references might be called `ap_verify.py` rather than `ap_verify` and also link to this reference page.

doc/lsst.ap.verify/failsafe.rst
```diff
@@ -0,0 +1,32 @@
...
+If a task fails with a fatal error, ``ap_verify`` will clean up and shut down.
+In particular, where possible it will :ref:`preserve metrics<ap-verify-failsafe-partialmetric>` computed before the failure point.
+
+.. _ap-verify-failsafe-partialmetric:
+
+Recovering Metrics From Partial Runs
+------------------------------------
+
+``ap_verify`` produces some measurements even if the pipeline cannot run to completion.
+Specifically, if a task fails, any ``lsst.verify.Measurement`` objects created by previous top-level tasks will be :ref:`stored to disk<ap-verify-results>`.
```
jonathansick: Use single backticks to link to the API:

```rst
`lsst.verify.Measurement`
```

doc/lsst.ap.verify/running.rst
```diff
@@ -0,0 +1,80 @@
+.. _ap-verify-running:
+
+#######################################
+Running ap_verify From the Command Line
+#######################################
+
+``ap_verify`` is a Python script designed to be run on both developer machines and verification servers.
+This page describes the minimum options needed to run ``ap_verify``.
+For more details, see the :ref:`ap-verify-cmd` or run ``ap_verify`` with the ``-h`` argument.
```
jonathansick:
```rst
[...] or run  :option:`ap_verify.py -h`.
```

(links to the reference docs).

python/lsst/ap/verify/ap_verify.py
```diff
@@ -98,65 +98,75 @@ def __call__(self, value):
...
+def _measureFinalProperties(metadata, metricsJob):
     """Measure any metrics that apply to the final result of the AP pipeline,
     rather than to a particular processing stage.
 
     Parameters
     ----------
     metadata: `lsst.daf.base.PropertySet`
         The metadata produced by the AP pipeline.
-    metrics_job: `verify.Job`
+    metricsJob: `verify.Job`
```
jonathansick: `lsst.verify.Job`

— needs the actual namespace to make the link.

python/lsst/ap/verify/measurements/compute_metrics.py
```diff
@@ -28,17 +28,17 @@
...
 
 from lsst.ap.verify.config import Config
-from .profiling import measure_runtime
+from .profiling import measureRuntime
 
 
-def measure_from_metadata(metadata):
-    """Attempts to compute all known metrics on Task metadata.
+def measureFromMetadata(metadata):
+    """Attempt to compute all known metrics on Task metadata.
```
jonathansick: Leave out the "Attempt to" part.

python/lsst/ap/verify/measurements/compute_metrics.py
```diff
@@ -47,18 +47,19 @@ def measure_from_metadata(metadata):
...
     Returns
     -------
-    a list of `lsst.verify.Measurement` derived from `metadata`. May be empty.
+    a list of `lsst.verify.Measurement` derived from ``metadata``. May be empty.
 
     Raises
     ------
-    `RuntimeError`:
-        the config file exists, but does not contain the expected data
+    RuntimeError:
```
jonathansick: No colon after the exception in the Raises section

```rst
Raises
------
RuntimeError
    The ``ap_verify`` configuration file exists, but does not contain the
    expected data under ``measurements``.
```

(check the Raises sections of other docstrings)

python/lsst/ap/verify/measurements/profiling.py
```diff
@@ -27,44 +27,44 @@
...
-    metric_name: `str`
+    metricName: `str`
         The fully qualified name of the metric being measured, e.g.,
         "pipe_tasks.ProcessCcdTime"
 
     Returns
     -------
-    an `lsst.verify.Measurement` for `metric_name`, or `None` if the timing
-    information for `task_name` is not present in `metadata`
+    an `lsst.verify.Measurement` for `metricName`, or `None` if the timing
```
jonathansick: Returns is formatted like Parameters, even if you need to make up the name of the return variable. For example, I'd write this one as:

```rst
Returns
-------
measurement : `lsst.verify.Measurement`
    The metric measurement, or `None` if timing information for ``taskName`` is not present in the ``metadata``.
```

doc/ap_verify/index.rst
kfindeisen: So `:mod:` is also redundant?

doc/lsst.ap.verify/command-line-reference.rst
kfindeisen: Ok. I was using https://raw.githubusercontent.com/lsst/pipe_base/master/doc/lsst.pipe.base/command-line-task-argument-reference.rst as a template, which doesn't have that.

doc/lsst.ap.verify/command-line-reference.rst
kfindeisen: I'm not sure I understand. What exactly is the distinction (I think I almost always mean "piece of software" when I say "ap_verify"), and why would I want `ap_verify`'s documentation to link to itself?

doc/lsst.ap.verify/running.rst
kfindeisen: I disagree. I think trying to use compound dataIds is an easy enough mistake to make that users won't think to check the reference and read the warning there.

python/lsst/ap/verify/ap_verify.py
kfindeisen: Huh. It looks like the dev guide contradicts itself here:
* https://developer.lsst.io/docs/py_docs.html#parameter-types says to use the full namespace
* https://developer.lsst.io/docs/py_docs.html#see-also says to start with the lowest common ancestor

Does the link generation actually work differently in these two contexts?

doc/lsst.ap.verify/command-line-reference.rst
jonathansick: Yeah, maybe I could add `program`, but in that case there was no specific script being reference, so it seemed safe to have the flags default to pointing to that command-line task page rather than being namespaced.

doc/lsst.ap.verify/command-line-reference.rst
jonathansick: I say this because ``ap_verify`` the EUPS packages doesn't have command-line arguments, but ``ap_verify.py`` the command-line script does have arguments.

python/lsst/ap/verify/ap_verify.py
jonathansick: Ah, I may need to update those docs. I guess we'll see if the link works when it's integrated into pipelines.lsst.io.

kfindeisen: Related question: should exceptions also be fully qualified? (There are a few that are not in `__builtins__`.)

jonathansick: Yes, fully-qualified.

doc/lsst.ap.verify/running.rst
```diff
@@ -0,0 +1,80 @@
...
+How to Run ap_verify in a New Workspace
+---------------------------------------
+
+Using the :ref:`HiTS 2015 <ap_verify_hits2015-package>` dataset as an example, one can run ``ap_verify`` as follows:
+
+.. TODO: why dataIdString instead of id or dataId? (DM-12853)
+
+.. prompt:: bash
+
+   python ap_verify/bin/ap_verify.py --dataset HiTS2015 --output workspace/hits/ --dataIdString "visit=54123 ccd=25 filter=g" --silent
```
ebellm: I would note here that running `ap_verify` in parallel causes the metric output to be clobbered, and reference DM-13042 (which is the ticket to fix that).

doc/lsst.ap.verify/running.rst
```diff
@@ -0,0 +1,80 @@
...
+The :command:`--rerun run1` argument will create a workspace in :file:`<hits-data>/rerun/run1/`.
+Since datasets are :ref:`not, in general, repositories<ap-verify-datasets-butler>`, the :option:`--rerun <ap_verify.py --rerun>` parameter only superficially resembles the analogous argument for command-line tasks.
+In particular, ``ap_verify``'s ``--rerun`` does not support repository chaining (as in :command:`--rerun input:output`); the input for ``ap_verify`` will always be determined by the :option:`--dataset <ap_verify.py --dataset>`.
+
+.. _ap-verify-results:
+
+How to Use Measurements of Metrics
+----------------------------------
+
+After ``ap_verify`` has run, it will produce a file named :file:`ap_verify.verify.json` in the working directory.
```
ebellm: I would make the same comment here about parallelization also.

doc/lsst.ap.verify/running.rst
kfindeisen: Are you sure? Documentation doesn't normally mention bugs unless they're not going to be fixed for a while, and this one seems pretty straightforward.

kfindeisen:
A [compiled version](https://github.com/lsst-dm/ap_verify/files/1560953/docs.zip) [zip] of the latest docs.

doc/lsst.ap.verify/running.rst
```diff
@@ -0,0 +1,88 @@
...
+
+   The command-line interface for ``ap_verify`` is at present much more limited than those of command-line tasks.
+   In particular, only file-based repositories are supported, and compound dataIds cannot be provided.
+   See the :ref:`ap-verify-cmd` for details.
+
+.. TODO: remove this note after resolving DM-13042
+
+.. warning::
+
+   ``ap_verify.py`` is not designed to have multiple instances running concurrently.
```
ebellm: how about "...does not currently support running multiple instances concurrently."  instead.

kfindeisen: Sure.

PR #13, opened 2017-11-30
DM-11155: Implement object count metrics

morriscb:


python/lsst/ap/verify/ap_verify.py
```diff
@@ -137,7 +139,17 @@ def _measure_final_properties(metadata, metrics_job):
     metrics_job: `verify.Job`
         The Job object to which to add any metric measurements made.
     """
-    measurements = measure_from_metadata(metadata)
+    measurements = []
+    measurements.extend(measure_from_metadata(metadata))
+    # In the current version of ap_pipe, DIFFIM_DIR has a parent of
+    # PROCESSED_DIR. This means that a butler created from the DIFFIM_DIR reop
+    # includes data from PROCESSED_DIR.
+    print(args)
```
isullivan: The print statement looks like it's left over from debugging. Should this function always be verbose?

python/lsst/ap/verify/measurements/association.py
```diff
@@ -0,0 +1,224 @@
...
+    if not metadata.exists("association.numNewDiaObjects"):
+        return None
+
+    n_new = metadata.getAsInt("association.numNewDiaObjects")
+    meas = lsst.verify.Measurement(metric_name, n_new * u.count)
+    return meas
+
+
+def measure_number_unassociated_dia_objects(metadata, task_name, metric_name):
+    """ Computes the number previously created DIAObjects that do loaded but
```
isullivan: "that do loaded" - typo

python/lsst/ap/verify/measurements/compute_metrics.py
```diff
@@ -62,4 +74,84 @@ def measure_from_metadata(metadata):
...
+    if 'visit' not in dataId_dict.keys():
+        raise RuntimeError('The dataId string is missing \'visit\'')
+    else:
+        visit = int(dataId_dict['visit'])
+        dataId_dict['visit'] = visit
+    if 'ccdnum' not in dataId_dict.keys():
+        raise RuntimeError('The dataId string is missing \'ccdnum\'')
+    else:
+        ccdnum = int(dataId_dict['ccdnum'])
+        dataId_dict['ccdnum'] = ccdnum
```
isullivan: Since you have to hard code these values for now, you might consider putting all of this in a separate function, something like `check_dataId`. 

Also, if this is functionality you hope will change in the future, I suggest you also file a bug ticket and reference that here.

python/lsst/ap/verify/measurements/compute_metrics.py
```diff
@@ -62,4 +74,84 @@ def measure_from_metadata(metadata):
...
+
+    measurement = measure_fraction_dia_sources_to_sci_sources(
+        butler, dataId_dict, "ip_diffim.fracDiaSourcesToSciSources")
+    if measurement is not None:
+        result.append(measurement)
+    return result
+
+
+def measure_from_L1_db_sqlite(db_name):
+    """ Make measurements on a sqlite database containing associated DIAObjects
```
isullivan: Based on the code below, do you mean "unassociated DIAObjects"? If not, maybe add a comment in the code below explaining the measurement further.

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -311,3 +312,25 @@ def run_ap_pipe(dataset, working_repo, parsed_cmd_line, metrics_job):
     _post_process(working_repo)
     log.info('Pipeline complete')
     return metadata
+
+
+def _get_ap_pipe_repos(metadata):
+    """ Retrieve the subdirectories and reops defined in ap_pipe and store them
```
isullivan: "reops" - typo

tests/test_association.py
```diff
@@ -0,0 +1,357 @@
...
+                np.random.rand() * 360 * afwGeom.degrees,
+                np.random.rand() * scatter_arcsec * afwGeom.arcseconds)
+        src.setCoord(coord)
+
+    return sources
+
+
+class MeasureAssociationTestSuite(lsst.utils.tests.TestCase):
+
+    def setUp(self):
```
isullivan: Setting up a butler for tests can sometimes be very slow. If that's the case here, it might be better to use `setUpClass` instead for some of the setup.

tests/test_association.py
```diff
@@ -0,0 +1,357 @@
...
+        os.remove(self.db_file)
+
+    def test_valid_from_metadata(self):
+        """Verify that association information can be recovered from metadata.
+        """
+        # Insert data into the task metadata.
+        test_assoc_result = pipeBase.Struct(
+            n_updated_dia_objects=5,
+            n_new_dia_objects=6,
+            n_unassociated_dia_objects=7,)
```
isullivan: Could you set these three values as named variables first, and then reference those variables (times `u.count`) in the asserts below?

tests/test_association.py
```diff
@@ -0,0 +1,357 @@
...
+        """
+        meas = measure_number_sci_sources(
+            self.butler,
+            dataId_dict=dataId_dict,
+            metric_name='ip_diffim.numSciSrc')
+        self.assertIsInstance(meas, Measurement)
+        self.assertEqual(
+            meas.metric_name,
+            lsst.verify.Name(metric='ip_diffim.numSciSrc'))
+        self.assertEqual(meas.quantity, 10 * u.count)
```
isullivan: Where does the 10 come from?

tests/test_association.py
```diff
@@ -0,0 +1,357 @@
...
+
+        meas = measure_total_unassociated_dia_objects(
+            cursor,
+            metric_name='association.numTotalUnassociatedDiaObjects')
+        self.assertIsInstance(meas, Measurement)
+        self.assertEqual(
+            meas.metric_name,
+            lsst.verify.Name(
+                metric='association.numTotalUnassociatedDiaObjects'))
+        self.assertEqual(meas.quantity, 5 * u.count)
```
isullivan: Where does the 5 come from?

tests/test_association.py
```diff
@@ -0,0 +1,357 @@
...
+            measure_number_sci_sources(
+                self.butler,
+                dataId_dict={'visit': 1000, 'filter': 'r'},
+                metric_name='ip_diffim.fracDiaSrcToSciSrc')
+
+        with self.assertRaises(dafPersist.NoResults):
+            measure_fraction_dia_sources_to_sci_sources(
+                self.butler,
+                dataId_dict={'visit': 1000, 'filter': 'r'},
+                metric_name='ip_diffim.fracDiaSrcToSciSrc')
```
isullivan: You might also check that you get an error if you supply the correct visit number, but a missing filter.

tests/test_association.py
```diff
@@ -0,0 +1,357 @@
...
+            measure_fraction_updated_dia_objects(
+                metadata, "association", "foo.bar.FooBar")
+            
+        with self.assertRaises(TypeError):
+            measure_number_sci_sources(
+                self.butler, dataId={'visit': 1111, 'filter': 'r'},
+                metric_name='foo.bar.FooBar')
+        with self.assertRaises(TypeError):
+            measure_fraction_dia_sources_to_sci_sources(
+                self.butler, dataId={'visit': 1111, 'filter': 'r'},
```
isullivan: I'd suggest you use your `dataId_dict` here instead of hardcoding the values.

python/lsst/ap/verify/measurements/compute_metrics.py
morriscb: Added a new function _convert_data_id_string

python/lsst/ap/verify/measurements/compute_metrics.py
morriscb: The idea here is that this is a general use method for computing any measurements that need to access the data products in the association database. I've changed the docstring to this "Make measurements on a sqlite database containing the results of from source association."

tests/test_association.py
morriscb: It doesn't seem to be currently. The tests take overall 2 seconds.

tests/test_association.py
morriscb: Changed this to a variable stored in self.

tests/test_association.py
morriscb: Same here.

PR #14, opened 2017-12-04
DM-12535: Wrap ap_verify and run it over HITS dataset

kfindeisen:
This PR adds a wrapper script that calls `ap_verify` for a specific set of visits and CCDs. The script has the data IDs hard-coded (since I couldn't see another way to extract them pre-ingestion), and has highly special-cased handling of measurement merging. Once DM-12314 is resolved, the script will be unneccessary, as `ap_verify`/`ap_pipe` will be able to internally handle multiple dataIds.

### DIFF
--- a/demo_run.py
+++ b/demo_run.py
@@ -1,0 +1,301 @@
+#
+# LSST Data Management System
+#
+# Copyright 2008-2017  AURA/LSST.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+
+"""Run ap_verify on HiTS data
+This is a temporary script, to be removed once ap_pipe can handle multiple dataIds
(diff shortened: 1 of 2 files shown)


python/lsst/ap/verify/ap_verify.py
```diff
@@ -153,11 +153,19 @@ def _measure_final_properties(metadata, output_dir, args, metrics_job):
         metrics_job.measurements.insert(measurement)
 
 
-def run_ap_verify():
+def run_ap_verify(cmdLine=None):
```
ebellm: is there a standard way to do this in LSST?  It seems unconventional...

kfindeisen: There's [CmdLineTask.parseAndRun](https://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/classlsst_1_1pipe_1_1base_1_1cmd_line_task_1_1_cmd_line_task.html#a5820e5ed05ca70b8e58efa8ee0567cdf)...

demo_run.py
```diff
@@ -0,0 +1,277 @@
...
+    """
+    if measurement is not None:
+        measurementSet.insert(measurement)
+
+
+def sumMeasurements(measurements):
+    """Adds the values of some measurements.
+
+    Extras and notes will be dictionary-merged (i.e., if multiple extras or
+    notes are assigned to the same key, only one will be preserved).
```
ebellm: which one?  the last one?

kfindeisen:
I'm working on the assumption that a lot of the merging code will not be needed at all in the final implementation (e.g., because metrics are only computed based on the final DB state). If it is needed, then it will certainly need to be more generic (see the discussion surrounding DMTN-057).

demo_run.py
kfindeisen: I'm not confident in guaranteeing that, especially since, if any extras or notes differ, this behavior is undesirable no matter which one gets chosen.

python/lsst/ap/verify/ap_verify.py
ebellm: Okay, part of the DM-11372 type work then.

PR #17, opened 2018-01-05
DM-12848: ap_verify unit tests fail on lsst-dev

kfindeisen:
This PR rewrites the unit tests for `lsst.ap.verify.measurements.profiling` to use a real task execution rather than a dummy one. This gives a running time greater than 0.01 seconds, fixing [DM-12848](https://jira.lsstcorp.org/browse/DM-12848). I also add a missing dependency on `ip_isr` to the package.

In a separate commit, I switch the `ups/*.cfg` and `ups/*.table` files from Windows to Unix line endings. I have no idea where those came from.


ups/ap_verify.cfg
```diff
@@ -1,12 +1,12 @@
...
+
+import lsst.sconsUtils
+
+dependencies = {
+}
+
+config = lsst.sconsUtils.Configuration(
+    __file__,
+    hasDoxygenInclude=False,
+    hasSwigFiles=False,
```
morriscb: I don't see anything that changed here. Is this just something weird the commit history?

kfindeisen: That's the "Fix Windows line endings" commit.

kfindeisen:
Thanks, I'll add some comments.

PR #18, opened 2018-01-09
DM-12306: Make datasets optional for ap_verify

kfindeisen:
This PR prevents unit tests from failing when the optional (data) dependency `ap_verify_hits2015` is not installed.


tests/test_dataset.py
```diff
@@ -27,9 +27,19 @@
...
 
+haveDataset = False
+try:
+    lsst.utils.getPackageDir("ap_verify_hits2015")
+    haveDataset = True
+except pexExcept.NotFoundError:
+    pass
+
+
+@unittest.skipIf(not haveDataset, "ap_verify_hits2015 not installed")
```
parejkoj: The other option is to do this in `setUpClass()`, as below. I'm not sure it matters which you use, although if you do need that packageDir in the class (e.g., for butler things), using setUpClass makes that easy. You should change the `not installed` to `not setup`, since it's the latter that really matters.

```
@classmethod
def setUpClass(cls)
        try:
            cls.dataDir = lsst.utils.getPackageDir("ap_verify_hits2015")
        except lsst.pex.exceptions.NotFoundError:
            raise unittest.SkipTest("ap_verify_hits2015 not setup")
```

kfindeisen: In this case it's "not installed", because the `Dataset` class sets up a dataset package during initialization. A bit different from our standard use of EUPS, I know.

kfindeisen:
`testInit` tests an important part of `Dataset`, but you're right that `ap_verify_hits2015` needs to not be set up in advance. That should be something I can add.

kfindeisen:
I'm rewriting the check to use `eups.Eups.findProduct` instead of `lsst.utils.getPackageDir`. I've found one bug in the `Dataset` constructor in the process. :grinning:

timj:
Please don't. We aren't meant to be using eups in code in general. It has to be an extremely good reason to have an eups dependency.

kfindeisen:
Ok. I'll remove self-setup as part of the changes on DM-12853, and leave the test disabler in a form that's forward-compatible with that.

PR #19, opened 2018-02-05
DM-12853: Review design questions in ap_verify

kfindeisen:
This PR updates the `ap_verify` documentation to comply with the decisions made on [DM-12853](https://jira.lsstcorp.org/browse/DM-12853). Two changes require concurrent updates to source code: renaming `--dataIdString` to `--id` and removing self-setup capability from datasets.

The built documentation is available in [zip format](https://github.com/lsst-dm/ap_verify/files/1696892/docs.zip).

### DIFF
--- a/python/lsst/ap/verify/dataset.py
+++ b/python/lsst/ap/verify/dataset.py
@@ -24,9 +24,9 @@ from __future__ import absolute_import, division, print_function
 
 import os
 import shutil
+from future.utils import raise_from
 
-from eups import Eups
-
+import lsst.pex.exceptions as pexExcept
 from lsst.utils import getPackageDir
 
 from .config import Config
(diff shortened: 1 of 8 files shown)


python/lsst/ap/verify/dataset.py
```diff
@@ -90,20 +90,27 @@ def __init__(self, datasetId):
         except KeyError:
             raise ValueError('Unsupported dataset: ' + datasetId)
 
-        self._dataRootDir = getPackageDir(datasetPackage)
-        self._validatePackage()
+        try:
+            self._dataRootDir = getPackageDir(datasetPackage)
+            self._validatePackage()
+        except pexExcept.NotFoundError as e:
```
jdswinbank: Can `_validatePackage` throw a `pexExcept.NotFoundError`? If it can (which I'm dubious about), would it mean that the package hasn't been set up, as the `RuntimeError` below claims?

timj: It's getPackageDir that is throwing the NotFoundError if the package is not set up.

jdswinbank: Yes, of course. Hence my comment indicate that there's no point having the `except` block around `_validatePackage`...

kfindeisen: I put them in the same `try` block because `self._dataRootDir` being initialized is a precondition for `_validatePackage`, and this way it's easier to see that the precondition will always be satisfied. As noted, `_validatePackage()` will never raise `NotFoundError`, so the exception handling is unambiguous.

timj: The usual recommendation is to use an `else` block with the `try`. Fluent Python makes an explicit case for doing that (the `else` only triggers if the try block completed).

kfindeisen: Fine, I'll do that.

PR #20, opened 2018-02-08
DM-13451: Make ap_verify responsible for ingestion

kfindeisen:
This PR moves ingestion out of `pipeline_driver` and into a new module called `ingestion`. The new module is responsible for the entire conversion of raws, calibs, and templates from the dataset format into a Butler repository, including some steps that were previously part of the main `ap_verify` module. The repository is then used for all processing by `pipeline_driver`.

Due to API changes, must be merged simultaneously with lsst-dm/ap_pipe#14.

### DIFF
--- a/python/lsst/ap/verify/ingestion.py
+++ b/python/lsst/ap/verify/ingestion.py
@@ -49,11 +49,11 @@ import lsst.daf.base as dafBase
 from lsst.ap.pipe import doIngestTemplates
 
 # Names of directories to be created in specified repository
-INGESTED_DIR = 'ingested'
-CALIBINGESTED_DIR = 'calibingested'
+_INGESTED_DIR = 'ingested'
+_CALIBINGESTED_DIR = 'calibingested'
 
 # Name of defects tarball residing in dataset's defects directory
-DEFECT_TARBALL = 'defects_2014-12-05.tar.gz'
+_DEFECT_TARBALL = 'defects_2014-12-05.tar.gz'
 
 
 def ingestDataset(dataset, repository):
@@ -67,8 +67,8 @@ def ingestDataset(dataset, repository):
         The dataset to be ingested.
     repository : `str`
         The file location of a repository to which the data will be ingested.
-        Shall be created if it does not exist. If `repository` does exist, it
-        must be compatible with `dataset` (in particular, it must support the
+        Shall be created if it does not exist. If ``repository`` does exist, it
+        must be compatible with ``dataset`` (in particular, it must support the
         relevant ``obs`` package).
 
     Returns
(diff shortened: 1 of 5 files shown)


python/lsst/ap/verify/ingestion.py
```diff
@@ -0,0 +1,480 @@
...
+
+    Parameters
+    ----------
+    dataset : `lsst.ap.verify.dataset.Dataset`
+        The dataset to be ingested.
+    repository : `str`
+        The file location of a repository to which the data will be ingested.
+        Shall be created if it does not exist. If ``repository`` does exist, it
+        must be compatible with ``dataset`` (in particular, it must support the
+        correct instrument).
```
mrawls: Not immediately clear what is meant by "it must support the correct instrument."

python/lsst/ap/verify/ingestion.py
```diff
@@ -0,0 +1,480 @@
...
+    metadata : `lsst.daf.base.PropertySet`
+        The full metadata from any Tasks called by this function, or `None`.
+    """
+    rawRepo = _getOutputRepo(workingRepo, _INGESTED_DIR)
+    dataFiles = _getDataFiles(dataset.rawLocation)
+    return _doIngest(rawRepo, dataset.refcatsLocation, dataFiles)
+
+
+def _ingestCalibs(dataset, workingRepo):
+    """Ingest the raw calibrations for use by LSST.
```
mrawls: Calling this "raw calibrations" is a little misleading for how they are used (we are ingesting DECam MasterCals from the community pipeline, not raw flats/biases that need to be combined before just being applied to the images). This isn't just a DECam quibble; this ingestion routine does not do anything to process the flats/biases and assumes they are ready to be applied to the raw science images as-is.

python/lsst/ap/verify/ingestion.py
```diff
@@ -0,0 +1,480 @@
...
+    workingRepo : `str`
+        The repository in which temporary products will be created. Must be
+        compatible with ``dataset``.
+
+    Returns
+    -------
+    metadata : `lsst.daf.base.PropertySet`
+        The full metadata from any Tasks called by this function, or `None`.
+    """
+    # TODO: move doIngestTemplates to this module once DM-11865 resolved
```
mrawls: I like how you've handled template "ingestion" for now, this is a good placeholder for post- DM-11865.

python/lsst/ap/verify/ingestion.py
```diff
@@ -0,0 +1,480 @@
...
+        Metadata from the `IngestCalibTask` (defects) for use by ``ap_verify``
+
+    Notes
+    -----
+    This function assumes very particular things about defect ingestion:
+    - They must live in a .tar.gz file in the same location on disk as the other calibs
+    - They will be ingested using ingestCalibs.py run from the ``calibRepo`` directory
+    - They will be manually uncompressed and saved in :file:`calibRepo/defects/<tarballname>/`.
+    - They will be added to the calib registry, but not linked like the flats and biases
+    """
```
mrawls: Would be good to add a TODO in here referencing DM-5467 (i.e., it is known that DECam defect ingestion is awful and there is an outstanding ticket to make it less awful)

python/lsst/ap/verify/ingestion.py
kfindeisen: I mean the way the `_mapper` needs to know about the obs package it's supposed to be using. Is there an implementation-independent way of saying that?

python/lsst/ap/verify/ingestion.py
kfindeisen: Per discussion with @jdswinbank, I think this is actually how we want `ap_verify` to handle templates in the long term. The AP pipeline will (probably) see the latest LSST data and the templates as a single repository, so we should try to simulate that.

mrawls: Interesting, OK!

python/lsst/ap/verify/ingestion.py
mrawls: Ah, I see. Perhaps just say "it must support the relevant obs_ package" ?

PR #23, opened 2018-03-16
DM-11162: Replace all use of Coord and subclasses with SpherePoint

r-owen:


tests/test_association.py
```diff
@@ -132,10 +131,10 @@ def createTestSources(nSources=5,
     for srcIdx in range(nSources):
         src = sources.addNew()
         src['id'] = srcIdx + startId
-        coord = Coord(sourceLocsDeg[srcIdx][0] * afwGeom.degrees,
-                      sourceLocsDeg[srcIdx][1] * afwGeom.degrees)
+        coord = afwGeom.SpherePoint(sourceLocsDeg[srcIdx][0],
+                                    sourceLocsDeg[srcIdx][1], afwGeom.degrees)
         if scatterArcsec > 0.0:
-            coord.offset(
+            coord = coord.offset(
```
kfindeisen: I suspect there need to be parallel changes in `ap_association`. See also [DM-11868](https://jira.lsstcorp.org/browse/DM-11868).

More generally, are we required to keep code up to date in `lsst-dm`? That seems like a difficult problem, especially when the packages in question are not part of CI.

r-owen: I already updated `ap_association` and `ap_pipe`. I was not required to do so, but I had them checked out and it was easy to do. If you like, I will leave merging of those three packages to you.

kfindeisen: No, it's fine. I must have missed the notification for `ap_association`.

PR #24, opened 2018-03-20
DM-13524: Add unit tests for ingestion

kfindeisen:
This PR adds minimalist tests of ingestion. A couple of minor bugs are fixed, but the rest are deferred to future tickets.

There is no test for attempts to ingest nonexistent files, as the default `*IngestionTask` behavior (log and ignore) is difficult to test.


tests/test_ingestion.py
```diff
@@ -0,0 +1,183 @@
...
+        # Making the directory appears to be both necessary and sufficient
+        os.mkdir(self._calibRepo)
+
+    def _rawButler(self):
+        """Return multiple ways of querying calibration repositories.
+
+        Returns
+        -------
+        butler : `lsst.daf.persistence.Butler`
+            A butler that should all be capable of finding ingested science data.
```
mrawls: Here and in the next one, "A butler that should all be capable of finding" is oddly worded. Do you mean "should be capable of finding all ingested [...] data" ?

tests/test_ingestion.py
```diff
@@ -0,0 +1,183 @@
...
+        butlers : `lsst.daf.persistence.Butler`
+            A butler that should all be capable of finding ingested calibration data.
+        """
+        return dafPersist.Butler(inputs={
+            'root': self._repo,
+            'mode': 'r',
+            'mapperArgs': {'calibRoot': self._calibRepo}})
+
+    def tearDown(self):
+        shutil.rmtree(self._repo, ignore_errors=True)
```
mrawls: This does make sense here but I'm just laughing to myself about `ignore_errors` as an argument 😆

tests/test_ingestion.py
```diff
@@ -0,0 +1,183 @@
...
+        """Test that ingesting calibrations adds them to a repository.
+        """
+        files = [os.path.join(IngestionTestSuite.testData,
+                              'rawData',
+                              'cpCalib',
+                              'masterCal',
+                              calibFile) for calibFile in
+                 [os.path.join('fci.fits'),
+                  os.path.join('zci.fits')]
+                 ]
```
mrawls: This definition of files is difficult to make sense of over several lines. Consider defining testDataDir first and then grabbing the two files.

tests/test_ingestion.py
```diff
@@ -0,0 +1,183 @@
...
+        ingestion._defectIngest(self._repo, self._calibRepo, files)
+
+        butler = self._calibButler()
+        self.assertTrue(butler.datasetExists('defects', dataId=IngestionTestSuite.defectDataId))
+
+    @unittest.skip("Ingestion functions cannot handle empty file lists, see DM-13835")
+    @unittest.skip("Dataset enumeration requires specific data keys for date, filter, etc., see DM-12762")
+    def testNoFileIngest(self):
+        """Test that attempts to ingest nothing do nothing.
+        """
```
mrawls: I'm not convinced that loudly crashing is a bad thing if a user tries to ingest nothing, because I can't think of a use case where I would intentionally want to ingest nothing. Am I missing something?

kfindeisen: I can't think of a specific use case either, but crashing over something like this is extremely unusual/surprising behavior. For example, if you try to iterate over an empty list, you expect the loop to simply do nothing instead of punishing you for not writing special-case code that would have amounted to the same thing.

Also, having a low-level function exit because of invalid inputs is extremely bad practice. If we want to ban empty file lists, the function should throw an exception so that the *program* can decide whether or not an exit is warranted (and have a chance to clean up properly either way).

PR #27, opened 2018-04-11
DM-13878: Confirm that ap_verify's documentation is sphinx-buildable

kfindeisen:
This PR brings `ap_verify` in sync with other Sphinx-documented packages. Fortunately, most of the changes that needed to be made were cosmetic.

Since the biggest change by line count was converting three files from Windows to Unix line endings, this PR may be best reviewed commit by commit.


kfindeisen:
The intent of documenting `lsst.ap.verify.measurements` separately was that people might end up making lots of changes to that package in order to support new metrics. When we were developing demo metrics for `ap_verify` I found that I had to explain the reasoning behind the package to keep people from re-coupling `measurements` to the main `ap_verify` module. The `automodapi` call is mostly a formality.

jonathansick:
Ah, got it.

PR #28, opened 2018-05-08
DM-14359: Fix data ID handling in ap_*

kfindeisen:
This PR cleans up the ingestion unit tests and guarantees that the pipeline driver always passes complete data IDs to `ApPipeTask`. It also removes use of `future` and other Python 2/3 compatibility code.

This PR must be merged after lsst/daf_persistence#93.

### DIFF
--- a/python/lsst/ap/verify/dataset.py
+++ b/python/lsst/ap/verify/dataset.py
@@ -21,10 +21,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-from __future__ import absolute_import, division, print_function
-
 import os
-from future.utils import raise_from
 
 from lsst.daf.persistence import Butler
 import lsst.pex.exceptions as pexExcept
@@ -64,10 +61,9 @@ class Dataset(object):
         try:
             self._dataRootDir = getPackageDir(datasetPackage)
         except pexExcept.NotFoundError as e:
-            raise_from(
-                RuntimeError('Dataset %s requires the %s package, which has not been set up.'
-                             % (datasetId, datasetPackage)),
-                e)
+            error = 'Dataset %s requires the %s package, which has not been set up.' \
+                % (datasetId, datasetPackage)
+            raise RuntimeError(error) from e
         else:
             self._validatePackage()
 
--- a/python/lsst/ap/verify/pipeline_driver.py
+++ b/python/lsst/ap/verify/pipeline_driver.py
@@ -29,14 +29,11 @@ that a total pipeline failure still allows some measurements to be
 recovered.
 """
 
-from __future__ import absolute_import, division, print_function
-
 __all__ = ["ApPipeParser", "MeasurementStorageError", "runApPipe"]
 
 import argparse
 import os
 import re
-from future.utils import raise_from
 
 import json
 
@@ -103,9 +100,7 @@ def _updateMetrics(metadata, job):
                 taskJob = Job.deserialize(**json.load(f))
             job += taskJob
     except (IOError, TypeError) as e:
-        raise_from(
-            MeasurementStorageError('Task metadata could not be read; possible downstream bug'),
-            e)
+        raise MeasurementStorageError('Task metadata could not be read; possible downstream bug') from e
 
 
 def _process(pipeline, workspace, dataId, parallelization):
@@ -235,13 +230,9 @@ def _deStringDataId(dataId):
     dataId: `dict` from `str` to any
         The dataId to be cleaned up.
     '''
-    try:
-        basestring
-    except NameError:
-        basestring = str
     integer = re.compile('^\s*[+-]?\d+\s*$')
     for key, value in dataId.items():
-        if isinstance(value, basestring) and integer.match(value) is not None:
+        if isinstance(value, str) and integer.match(value) is not None:
             dataId[key] = int(value)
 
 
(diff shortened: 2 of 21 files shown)


python/lsst/ap/verify/dataset.py
```diff
@@ -64,10 +61,9 @@ def __init__(self, datasetId):
         try:
             self._dataRootDir = getPackageDir(datasetPackage)
         except pexExcept.NotFoundError as e:
-            raise_from(
-                RuntimeError('Dataset %s requires the %s package, which has not been set up.'
-                             % (datasetId, datasetPackage)),
-                e)
+            raise RuntimeError('Dataset %s requires the %s package, which has not been set up.'
+                               % (datasetId, datasetPackage)) \
+                from e
```
parejkoj: Ok, `raise from` is neat and I was not aware of it. Thanks for making me aware!

Although, why can't you just put the `from e` immediately after the parentheses, instead of on a new line?

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -225,6 +220,53 @@ def runApPipe(metricsJob, workspace, parsedCmdLine):
...
+        Butler identifier naming the data to be processed by the underlying
+        task(s).
+
+    Returns
+    -------
+    dataRefs : iterable of `lsst.daf.persistence.ButlerDataRef`
+        All data references compatible with ``dataId``. The references shall be fully expanded.
+    """
+    refSuperset = butler.subset(datasetType=datasetType, dataId=dataId)
+    return [ref for ref in refSuperset if _dataExists(butler=butler, dataRef=ref)]
```
parejkoj: Ah, yes, this bit of code. I wonder if we shouldn't lift this up into the butler itself, since I have my own versions of it floating around in jointcal (e.g. https://github.com/lsst/jointcal/blob/master/bin.src/plot_jointcal_results.py#L20 ), and I bet others do to.

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -225,6 +220,53 @@ def runApPipe(metricsJob, workspace, parsedCmdLine):
...
+        `True` if data exists, `False` otherwise.
+    """
+    subItems = dataRef.subItems()
+    if subItems:
+        for subDR in subItems:
+            if _dataExists(butler, subDR):
+                return True
+        return False
+    else:
+        return dataRef.datasetExists()
```
parejkoj: This is another bit of code that should probably live in the butler itself.

tests/test_ingestion.py
```diff
@@ -146,7 +144,6 @@ def testDefectIngest(self):
         self.assertTrue(butler.datasetExists('defects', dataId=IngestionTestSuite.defectDataId))
 
     @unittest.skip("Ingestion functions cannot handle empty file lists, see DM-13835")
-    @unittest.skip("Dataset enumeration requires specific data keys for date, filter, etc., see DM-12762")
```
parejkoj: Yay removing skips!

tests/test_ingestion.py
```diff
@@ -184,6 +178,16 @@ def testFindMatchingFiles(self):
...
+
+    Parameters
+    ----------
+    datasetType : `str`
+        The type of dataset to search for.
+
+    Notes
+    -----
+    .. warning::
+       Does not work for calib datasets, because they're not discoverable.
```
parejkoj: Good note. Does this deserve a ticket and TODO?

python/lsst/ap/verify/dataset.py
kfindeisen: Because then I'm splitting a constructor argument across multiple lines, yet putting part of the argument and external logic on the same line? It seems less readable to me, but I won't claim to be an expert on the arcane rules of Python indentation.

python/lsst/ap/verify/pipeline_driver.py
kfindeisen: TBH, this is a near-duplicate of some code in [`pipe.base.ArgumentParser`](https://github.com/lsst/pipe_base/blob/master/python/lsst/pipe/base/argumentParser.py#L162-L166). I considered factoring it out, but `pipe_base` isn't really a natural place for this functionality. Your suggestion is better, but we'd have to *create* a Butler utilities module first.

tests/test_ingestion.py
kfindeisen: See discussion on [#dm](https://lsstc.slack.com/archives/C2JPL2DGD/p1525816663000392). To me, it felt like a "that's not a bug, that's a feature" moment.

python/lsst/ap/verify/pipeline_driver.py
parejkoj: It figures that a couple years after killing off `daf_butlerUtils`, I end up resurrecting it...

Why can't it just live in the butler itself though? `butler.dataSetReallyExistsIPromise()`?

python/lsst/ap/verify/dataset.py
kfindeisen: Yes there are -- `flake8` complains like crazy unless I get the continuation indent *exactly* right.

python/lsst/ap/verify/pipeline_driver.py
kfindeisen: To me it seems like secondary functionality (beware cluttering the API with convenience methods, etc.).

Also, IIRC Butler 2 was specifically designed to not worry about existence checks, as an optimization or something (plus it can be useful to have references to data that don't exist at the time the reference is constructed).

parejkoj: Thanks for moving it into daf_persistence.

PR #30, opened 2018-05-17
DM-14496: test_association broken

morriscb:


morriscb:
I use them in some of the tests for ap_association. The function there is a direct copy-paste from those tests. I can remove those variables if you would like.

PR #31, opened 2018-05-18
DM-13849: Convert all ap_verify test data to obs_test

kfindeisen:
This PR modifies `test_ingestion` and `test_dataset` to depend on `obs_test` rather than `obs_decam`, and fixes some more minor issues I came across while working on these tests.

This PR must be merged after lsst-dm/ap_verify_testdata#1.


tests/test_dataset.py
```diff
@@ -28,35 +28,35 @@
...
+        cls.obsPackage = 'obs_test'
+        cls.camera = 'test'
         try:
             lsst.utils.getPackageDir(cls.testDataset)
         except pexExcept.NotFoundError:
             raise unittest.SkipTest(cls.testDataset + ' not set up')
 
+        # Hack the config for testing purposes
+        # Note that Config.instance is supposed to be immutable, so some code may break
+        Config.instance._allInfo['datasets.' + cls.datasetKey] = cls.testDataset
```
mrawls: Your comments here don't exactly inspire confidence. What's "some code may break" all about?

python/lsst/ap/verify/dataset.py
```diff
@@ -53,9 +53,8 @@ class Dataset(object):
     """
 
     def __init__(self, datasetId):
-        try:
-            datasetPackage = self._getDatasetInfo()[datasetId]
-        except KeyError:
+        datasetPackage = self._getDatasetInfo()[datasetId]
```
mrawls: You're reasonably confident that trying to get a non-existent dataset will return None and won't fail in some other way?

kfindeisen: This is an apparently deliberate deviation of `daf.persistence.Policy` from `dict`-like behavior. 😞 

Since it's undocumented behavior, I guess I could try to handle *both* `None` return and `KeyError`, though that would involve duplicating code.

mrawls: I see. I lean toward whatever will make error messages less cryptic to users, which may mean duplicating code in this case.

tests/test_dataset.py
kfindeisen: `Config.instance` loads itself from `ap_verify`'s config files before it gets used (at import time, I think?). The assumption is that `Config.instance` simply reflects the contents of those files, i.e., it does not change within a program run.

So you could imagine that some code (yet to be written) reads something from `Config.instance` before `DatasetTestSuite.setUpClass` runs, then `Config.instance` gets modified, then the code reads `Config.instance` later and gets an inconsistent result, which causes it to do something crazy (kind of literally). This could be code in or called by a different test, so it's not possible to control it.

The alternative is actually putting `ap_verify_testdata` in the config. However, it would then be usable by `ap_verify` itself, and even show up as one of the available datasets in the `-h` text. A hack that only affects the tests but not the program itself seems like the lesser evil.

mrawls: Huh. That is a somewhat interesting hypothetical case! I didn't have the context that `Config.instance` isn't "normally" accessed from a place like this. Perhaps instead of "so some code may break" you could say "so if the config has changed since import there may be a discrepancy."

kfindeisen: Not sure what you mean by "normally accessed" -- any part of the `ap_verify` codebase is free to access `Config.instance` at will; that's what it's for.

How about:
> Note that Config.instance is supposed to be immutable, so, depending on initialization order, this modification may cause other tests to see inconsistent config values

PR #33, opened 2018-05-30
DM-13835: Cannot ingest empty data

kfindeisen:
This PR changes the behavior of `DatasetIngestTask`'s private methods from crashing to raising a specific exception. It also adds a unit test that could not be implemented because of this bug (at least, not while we were still using `testdata_decam`).

### DIFF
--- a/python/lsst/ap/verify/ingestion.py
+++ b/python/lsst/ap/verify/ingestion.py
@@ -438,7 +438,11 @@ def _runIngestTask(task, args):
         The command-line arguments for ``task``. Must be compatible with ``task.ArgumentParser``.
     """
     argumentParser = task.ArgumentParser(name=task.getName())
-    parsedCmd = argumentParser.parse_args(config=task.config, args=args)
+    try:
+        parsedCmd = argumentParser.parse_args(config=task.config, args=args)
+    except SystemExit as e:
+        # SystemExit is not an appropriate response when the arguments aren't user-supplied
+        raise ValueError("Invalid ingestion arguments: %s" % args) from e
     task.run(parsedCmd)
 
 
--- a/python/lsst/ap/verify/ingestion.py
+++ b/python/lsst/ap/verify/ingestion.py
@@ -31,6 +31,7 @@ __all__ = ["DatasetIngestConfig", "ingestDataset"]
 
 import fnmatch
 import os
+import pathlib
 import tarfile
 from glob import glob
 import sqlite3
@@ -166,10 +167,8 @@ class DatasetIngestTask(pipeBase.Task):
             ``obs`` package as this task's subtasks.
         """
         dataset.makeCompatibleRepo(workspace.dataRepo)
-        if not os.path.isdir(workspace.calibRepo):
-            os.mkdir(workspace.calibRepo)
-        if not os.path.isdir(workspace.templateRepo):
-            os.mkdir(workspace.templateRepo)
+        pathlib.Path(workspace.calibRepo).mkdir(parents=True, exist_ok=True)
+        pathlib.Path(workspace.templateRepo).mkdir(parents=True, exist_ok=True)
 
     def _ingestRaws(self, dataset, workspace):
         """Ingest the science data for use by LSST.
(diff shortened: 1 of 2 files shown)


python/lsst/ap/verify/ingestion.py
```diff
@@ -184,14 +184,22 @@ def _ingestRaws(self, dataset, workspace):
             The dataset on which the pipeline will be run.
         workspace : `lsst.ap.verify.workspace.Workspace`
             The location containing all ingestion repositories.
+
+        Raises
+        ------
+        RuntimeError
+            Raised if there are no files to ingest.
```
parejkoj: Thanks for documenting this. We need more docstrings that call out known exceptions.

python/lsst/ap/verify/ingestion.py
```diff
@@ -207,7 +215,15 @@ def _doIngest(self, repo, dataFiles, badFiles):
...
             A list of filenames to exclude from ingestion. Must not contain paths.
             May contain wildcards.
+
+        Raises
+        ------
+        RuntimeError
+            Raised if ``dataFiles`` is empty.
         """
+        if not dataFiles:
+            raise RuntimeError("No raw files to ingest.")
```
parejkoj: Maybe be a little more explicit here, to help with debugging: `"No raw files to ingest; dataFiles is %s."%dataFiles`

python/lsst/ap/verify/ingestion.py
```diff
@@ -256,7 +280,15 @@ def _doIngestCalibs(self, repo, calibRepo, calibDataFiles):
...
             but may include flats, biases, darks, fringes, or sky. May contain
             wildcards.
+
+        Raises
+        ------
+        RuntimeError
+            Raised if ``calibDataFiles`` is empty.
         """
+        if not calibDataFiles:
+            raise RuntimeError("No calib files to ingest.")
```
parejkoj: As above, please include `calibDataFiles` in the raised string.

python/lsst/ap/verify/ingestion.py
```diff
@@ -307,12 +344,21 @@ def _doIngestDefects(self, repo, calibRepo, defectTarball):
...
         """
         # TODO: clean up implementation after DM-5467 resolved
         defectDir = os.path.join(calibRepo, "defects")
-        if not os.path.isdir(defectDir):
-            os.mkdir(defectDir)
-        tarfile.open(defectTarball, "r").extractall(defectDir)
+        with tarfile.open(defectTarball, "r") as opened:
+            if opened.getNames():
+                if not os.path.isdir(defectDir):
+                    os.mkdir(defectDir)
```
parejkoj: Maybe use pathlib here in place of `if exist: mkdir`, to avoid `os.mkdir` race conditions on parallel ingest:

`pathlib.Path(defectDir).mkdir(parents=True, exist_ok=True) `

https://stackoverflow.com/a/14364249
https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

tests/test_ingestion.py
```diff
@@ -122,34 +127,43 @@ def testDataIngest(self):
...
+        badFiles = ['raw_v2_fg.fits.gz']
+
+        testDir = os.path.join(IngestionTestSuite.testData, 'raw')
+        files = [os.path.join(testDir, datum['file']) for datum in IngestionTestSuite.rawData]
+        self._task._doIngest(self._repo, files, badFiles)
+
+        butler = self._rawButler()
+        for datum in IngestionTestSuite.rawData:
+            dataId = {'visit': datum['visit']}
+            self.assertEqual(butler.datasetExists('raw', dataId), datum['file'] not in badFiles)
```
parejkoj: Ooog. I'm not too keen on this syntax (it was hard for me to parse), but I'm not sure of a better approach given the butler methods that exist.

python/lsst/ap/verify/ingestion.py
kfindeisen: The `if` statement guarantees that won't display anything interesting.

tests/test_ingestion.py
kfindeisen: I could do the approach of `if in badFiles: self.assertFalse(...) else: self.assertTrue(...)`, but I'd find that ickier.

python/lsst/ap/verify/ingestion.py
kfindeisen: I'm not sure parallel ingestion is possible, but your solution is more elegant.

python/lsst/ap/verify/ingestion.py
parejkoj: `dataFiles` could be `None`, `False`, `[]`, `()`, `{}`, `''`. It might be informative to know which of those it was.

kfindeisen: Ugh, weakly typed languages. 😞 I'll write it with `%r`, then.

PR #40, opened 2018-07-26
DM-15142: Create CI-sized DECam Dataset

kfindeisen:
This PR registers `ap_verify_ci_hits2015` with `ap_verify` and adds it to the documentation's list of known datasets. It also updates dataset installation instructions in response to lsst/repos#37.

### DIFF
--- a/ups/ap_verify.table
+++ b/ups/ap_verify.table
@@ -9,8 +9,11 @@ setupRequired(verify)
 setupRequired(ap_pipe)
 
 setupOptional(ap_verify_testdata)
-setupOptional(ap_verify_hits2015)
 setupOptional(obs_test)
+
+# Datasets other than ap_verify_testdata must be set up manually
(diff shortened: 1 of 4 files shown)


ups/ap_verify.table
```diff
@@ -9,8 +9,12 @@ setupRequired(verify)
 setupRequired(ap_pipe)
 
 setupOptional(ap_verify_testdata)
-setupOptional(ap_verify_hits2015)
 setupOptional(obs_test)
 
+# Datasets other than ap_verify_testdata must be set up manually
+#     to avoid being inadvertently downloaded by lsstsw
+# setupOptional(ap_verify_hits2015)
```
SimonKrughoff: I don't like having executable code commented out.  Maybe instead just list the packages you could use?  I.e.
`# setupOptional(ap_verify_hits2015)` --> `#      ap_verify_hits2015`

PR #46, opened 2018-09-19
DM-15742: Write shell script to serve as a Jenkins hook

kfindeisen:
This PR introduces a script, `run_ci_dataset.sh`, which encapsulates the environment setup and data ID handling needed to run `ap_verify` on an entire dataset. By analogy with `validate_drp/examples/processData.sh`, the script assumes that the dataset(s) to be run are already downloaded and EUPS set up, and that uploading metric files is the responsibility of the `.groovy` script.

The first commit fixes a fatal bug in `ap_verify` that was, ironically, missed due to lack of CI.

This PR cannot be merged before #40, as the script relies on the latter PR's definition of `CI-HiTS2015`.

### DIFF
--- a/bin/run_ci_dataset.sh
+++ b/bin/run_ci_dataset.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+
+# This file is part of ap_verify.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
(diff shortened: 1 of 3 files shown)


jhoblitt:
GPLv3 does not require a copyright notice per file.  It encourages it in the "How to Apply These Terms to Your New Programs" section, which isn't part of the terms of the license.  It does, however, forbid the removal of the boilerplate notices from derivative works.  I have a strong preference to limit source files to technical matters and to use a simple `LICENSE` file at the repo/project root.

/cc @jonathansick

timj:
@jhoblitt the developer guide defines the policy for license and copyright.

jonathansick:
@jhoblitt Sorry, I had a brain freeze yesterday and was thinking of an earlier draft version of the policy that did minimize the boilerplate.

jhoblitt:
This statement in the dev guide is technically incorrect:

> The GPL-3.0 license requires each source file to have a preamble comment containing a license statement. This is the generic license preamble:

If it is required, it is only by DM policy.

timj:
Gnu *strongly* prefer that the license preamble is in the file.  Why would we go against their wishes? I'm fine with you changing the dev guide to stop saying "requires" and instead say "strongly encourages" and that DM policy is to follow their advice.

jhoblitt:
Because it liters every file with boilerplate, many of which probably don't even qualify as a copyrightable work?  GNU wants people to use https://www.gnu.org/prep/standards/standards.html as well; not everything GNU is a good idea.

jdswinbank:
Hey folks, I suggest GitHub isn't a great venue for this discussion. @jhoblitt, if you'd like to suggest a change to the rules, please file a ticket on Jira. Thanks!

kfindeisen:
This conversation has derailed a bit from my (admittedly implied) question: does a shell script count as a "source file" for the purposes of DM's copyright policy?

jonathansick:
@kfindeisen Yep.

PR #51, opened 2018-11-08
DM-16319: ap_verify source count metrics do not exist

kfindeisen:
This PR changes the namespace of `ap_association` metrics from `association` to `ap_association`. This change is needed to correctly interoperate with the metrics as registered in lsst/verify_metrics#14.


afausti:
Looks good. I've tested this branch with SQuaSH and the new metrics were created correctly.

PR #54, opened 2018-11-27
DM-16333: Add Job metadata to the ap job.

SimonKrughoff:
@kfindeisen and @jhoblitt I think we could do something like this to get the appropriate metadata in the job.  Does this look reasonable, and how should I test?

### DIFF
--- a/python/lsst/ap/verify/pipeline_driver.py
+++ b/python/lsst/ap/verify/pipeline_driver.py
@@ -144,7 +144,6 @@ def runApPipe(metricsJob, workspace, parsedCmdLine):
     #  Insert job metadata including dataId
     metricsJob.meta.update({'instrument': _extract_instrument_from_butler(workspace.workButler)})
     metricsJob.meta.update(dataId)
-    #  I think we could probably send in the test repo directory too, but I don't know how to get that here.
 
     pipeline = apPipe.ApPipeTask(workspace.workButler, config=_getConfig(workspace))
     try:
--- a/python/lsst/ap/verify/pipeline_driver.py
+++ b/python/lsst/ap/verify/pipeline_driver.py
@@ -43,6 +43,19 @@ import lsst.ap.pipe as apPipe
 from lsst.verify import Job
 
 
+# borrowed from validate_drp
+def _extract_instrument_from_butler(butler):
+    """Extract the last part of the mapper name from a Butler repo.
+    'lsst.obs.lsstSim.lsstSimMapper.LsstSimMapper' -> 'LSSTSIM'
+    'lsst.obs.cfht.megacamMapper.MegacamMapper' -> 'CFHT'
+    'lsst.obs.decam.decamMapper.DecamMapper' -> 'DECAM'
+    'lsst.obs.hsc.hscMapper.HscMapper' -> 'HSC'
+    """
+    camera = butler.get('camera')
+    instrument = camera.getName()
+    return instrument.upper()
+
+
 class ApPipeParser(argparse.ArgumentParser):
     """An argument parser for data needed by ``ap_pipe`` activities.
 
@@ -125,6 +138,13 @@ def runApPipe(metricsJob, workspace, parsedCmdLine):
     log = lsst.log.Log.getLogger('ap.verify.pipeline_driver.runApPipe')
 
     dataId = _parseDataId(parsedCmdLine.dataId)
+    #  After processes are implemented, remove the flake exception
+    processes = parsedCmdLine.processes  # noqa: F841
+
+    #  Insert job metadata including dataId
+    metricsJob.meta.update({'instrument': _extract_instrument_from_butler(workspace.workButler)})
+    metricsJob.meta.update(dataId)
+    #  I think we could probably send in the test repo directory too, but I don't know how to get that here.
 
     pipeline = apPipe.ApPipeTask(workspace.workButler, config=_getConfig(workspace))
     try:
--- a/ups/ap_verify.table
+++ b/ups/ap_verify.table
@@ -7,6 +7,7 @@ setupRequired(ip_isr)
 setupRequired(pipe_tasks)
 setupRequired(verify)
 setupRequired(ap_pipe)
+setupRequired(ap_association)
 
 setupOptional(ap_verify_testdata)
 setupOptional(obs_test)


python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -202,6 +202,23 @@ def runApPipe(metricsJob, workspace, parsedCmdLine):
 
     dataId = _parseDataId(parsedCmdLine.dataId)
     processes = parsedCmdLine.processes
+    
+    # borrowed from validate_drp
+    def extract_instrument_from_butler(butler):
```
kfindeisen: Please make this a package-private function outside `runApPipe`. The implementation of `runApPipe` will be changing quite a bit in the near future, and the changes will be easier to merge if I don't have to code around a nested function definition.

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -202,6 +202,23 @@ def runApPipe(metricsJob, workspace, parsedCmdLine):
...
+        'lsst.obs.decam.decamMapper.DecamMapper' -> 'DECAM'
+        'lsst.obs.hsc.hscMapper.HscMapper' -> 'HSC'
+        """
+        camera = butler.get('camera')
+        instrument = camera.getName()
+        return instrument.upper()
+    
+    #  Insert job metadata
+    metricsJob.meta.update({'instrument': extract_instrument_from_butler(workspace.workButler),
+                            'filter_name': dataId['filter']})
```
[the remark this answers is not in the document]
SimonKrughoff: I think requires is too strong.  There is an assumption inherited from the `validate_drp` work, but the verify framework certainly doesn't require this.  It's useful metadata to have.  @afausti can say more, but I think the upcoming visualization framework will make this assumption go away.

python/lsst/ap/verify/pipeline_driver.py
SimonKrughoff: Will do.  Good point.

SimonKrughoff:
I think in general, we will want to associate the data ids involved in measuring a metric with the metric.  That's why I haven't considered including the filter name in the metadata as onerous when it makes sense.

SimonKrughoff:
OK.  I'll make the updates and try to test it out.  @jhoblitt should we set up a test env to test the branch?

python/lsst/ap/verify/pipeline_driver.py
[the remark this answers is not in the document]
kfindeisen: Ok, so until the old framework goes away, how should we handle metrics that can't be meaningfully associated with a filter?

SimonKrughoff:
@kfindeisen Have a look at this updated PR.  I've taken into account your comments.  Note that we no longer select specific metadata from the dataId, but ship the whole thing assuming it will be useful.  Also note that I also had to fix some linting errors.  I rebased yesterday, but some of these may still have been fixed.  I'll make sure to rebase on master before merging.

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -43,6 +43,19 @@
 from lsst.verify import Job
 
 
+# borrowed from validate_drp
+def _extract_instrument_from_butler(butler):
```
kfindeisen: :+1:

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -125,6 +138,13 @@ def runApPipe(metricsJob, workspace, parsedCmdLine):
     log = lsst.log.Log.getLogger('ap.verify.pipeline_driver.runApPipe')
 
     dataId = _parseDataId(parsedCmdLine.dataId)
+    #  After processes are implemented, remove the flake exception
+    processes = parsedCmdLine.processes  # noqa: F841
```
kfindeisen: I don't understand why you need these two lines at all.

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -125,6 +138,13 @@ def runApPipe(metricsJob, workspace, parsedCmdLine):
...
+
+    #  Insert job metadata including dataId
+    metricsJob.meta.update({'instrument': _extract_instrument_from_butler(workspace.workButler)})
+    metricsJob.meta.update(dataId)
```
kfindeisen: I assume this is the same tag we discussed with @afausti on DM-11321. Is it supposed to be job-level or measurement-level?

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -125,6 +138,13 @@ def runApPipe(metricsJob, workspace, parsedCmdLine):
...
+    #  I think we could probably send in the test repo directory too, but I don't know how to get that here.
```
kfindeisen: What do you mean by "test repo directory"? If it's something from the CI context, then it would cause us trouble if we tried to do a SQuaSH upload from elsewhere.

tests/test_dataset.py
```diff
@@ -59,10 +61,10 @@ def testDirectories(self):
...
         """
         root = self._testbed.datasetRoot
-        self.assertEqual(self._testbed.rawLocation, os.path.join(root, 'raw'))
-        self.assertEqual(self._testbed.calibLocation, os.path.join(root, 'calib'))
-        self.assertEqual(self._testbed.templateLocation, os.path.join(root, 'templates'))
-        self.assertEqual(self._testbed.refcatsLocation, os.path.join(root, 'refcats'))
+        self.assertEqual(self._testbed.rawLocation, join(root, 'raw'))
+        self.assertEqual(self._testbed.calibLocation, join(root, 'calib'))
+        self.assertEqual(self._testbed.templateLocation, join(root, 'templates'))
+        self.assertEqual(self._testbed.refcatsLocation, join(root, 'refcats'))
```
kfindeisen: Would it be possible to revert the changes to this file? I don't see how they relate to the new metadata, and IMO `join` is much more confusing than `os.path.join` (at first glance it looks like a straight string join).

python/lsst/ap/verify/pipeline_driver.py
kfindeisen: Also, it might be worth a comment that it's `update(dataId)` because a `daf.persistence.DataId` is not serializable... just got bitten by that myself!

python/lsst/ap/verify/pipeline_driver.py
SimonKrughoff: I didn't add them.  It was in there as a place holder (I assume), but because `processes` was not used, it failed linting.  I added the noqa so that flake8 would ignore that line until the `processes` functionality is implemented.

tests/test_dataset.py
SimonKrughoff: More linting problems.  It was really hard to get these under the line limit.  I can't simply revert because travis fails as it was.

python/lsst/ap/verify/pipeline_driver.py
SimonKrughoff: I'll remove the comment.

python/lsst/ap/verify/pipeline_driver.py
SimonKrughoff: It needs to be at Job level since the measurements are all ingested as a Job.  We could also record measurement level metadata, but that is out of scope for this.  As to the second comment, that is not why I used `update`.  I used it because that's the recommended way of merging dictionaries as far as I know.  If you have text for a comment, I'd be happy to add it, but I'm not sure I could come up with something on my own.

python/lsst/ap/verify/pipeline_driver.py
kfindeisen: Aha, it was removed in 0c57f674f4ccd5e22157ad0aaf67679b31ed7203, it must have gotten re-added during rebasing.

tests/test_dataset.py
kfindeisen: That can't be right. I've been building that code with Travis for months, and `setup.cfg` uses the science pipelines limit (110 characters).

python/lsst/ap/verify/pipeline_driver.py
kfindeisen: I'm not used to thinking of either `metricsJob.meta` or a data ID as a dictionary, but ok. As long as we're all on the same page about how and where the information should be provided.

python/lsst/ap/verify/pipeline_driver.py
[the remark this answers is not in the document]
SimonKrughoff: To bring this to a close, we have decided to deprecate the old framework to avoid this problem.  This is not an issue since there are no historical AP metrics in the old framework and would have only been populated for symmetry with DRP going forward.

tests/test_dataset.py
kfindeisen: Discussed this on Slack, but for the record: the only Travis errors I could find for this file were W504, which was fixed by #50. So it should be safe to drop this commit once that PR is in the history.

SimonKrughoff: OK, so travis has passed.

PR #57, opened 2018-12-13
DM-13887: Let ap_verify process multiple images per instance

kfindeisen:
This PR adds multiprocessing support to `ap_verify` by making the `pipeline_driver` module call `ApPipeTask.parseAndRun` instead of `ApPipeTask.runDataRef`. This change requires a complete reorganization of how `ap_verify` passes information from `ap_pipe` to the metrics code, but as a side benefit `ap_verify` no longer needs to parse data IDs itself.

### DIFF
--- a/bin/run_ci_dataset.sh
+++ b/bin/run_ci_dataset.sh
@@ -24,9 +24,8 @@
 # Simple script for running an entire dataset through ap_verify
 # Assumes that the requested dataset is already set up in EUPS
 #
-# The ap_verify workspace is a directory with the dataset name; until DM-13887
-# gets resolved, metrics are dumped to the workspace directory with names
-# containing each run's data ID.
+# The ap_verify workspace is a directory with the dataset name; metrics are
+# dumped to the workspace directory with names containing each run's data ID.
 
 set -e
 
@@ -71,7 +70,7 @@ fi
 # Would be created by ap_verify, but the OS might try to open the log first
 mkdir "${WORKSPACE}"
 
-# Store processor count (parellelism not yet implemented, see DM-13887)
+# Store processor count
 MACH=$(uname -s)
 if [[ $MACH == Darwin ]]; then
     sys_proc=$(sysctl -n hw.logicalcpu)
@@ -81,17 +80,10 @@ fi
 max_proc=8
 NUMPROC=${NUMPROC:-$((sys_proc < max_proc ? sys_proc : max_proc))}
 
-# Extract desired dataIds runs from config file
-# Workaround for DM-13887
-IDLIST="${PRODUCT_DIR}"/config/${DATASET}.list
-
-while read -r ID; do
-    echo "Running ap_verify on ${ID}..."
-    ap_verify.py --dataset "${DATASET}" \
-        --id "${ID}" \
-        --output "${WORKSPACE}" \
-        --processes "${NUMPROC}" \
-        --metrics-file "${WORKSPACE}/ap_verify.${ID}.verify.json" \
-        --silent \
-        &>> "${WORKSPACE}"/apVerify.log
-done < "${IDLIST}"
+echo "Running ap_verify on ${DATASET}..."
+ap_verify.py --dataset "${DATASET}" \
+    --output "${WORKSPACE}" \
+    --processes "${NUMPROC}" \
+    --metrics-file "${WORKSPACE}/ap_verify.{dataId}.verify.json" \
+    --silent \
+    &>> "${WORKSPACE}"/apVerify.log
--- a/doc/lsst.ap.verify/command-line-reference.rst
+++ b/doc/lsst.ap.verify/command-line-reference.rst
@@ -47,9 +47,6 @@ Required arguments are :option:`--dataset`, :option:`--id`, and :option:`--outpu
    Specify data ID to process using data ID syntax.
    For example, ``--id "visit=12345 ccd=1 filter=g"``.
    
-   Currently this argument is heavily restricted compared to its :doc:`command line task counterpart</modules/lsst.pipe.base/command-line-task-dataid-howto>`.
-   In particular, the dataId must specify exactly one visit and exactly one CCD, and may not be left blank to mean "all data".
-
 .. option:: --dataset <dataset_name>
 
    **Input dataset designation.**
@@ -73,10 +70,6 @@ Required arguments are :option:`--dataset`, :option:`--id`, and :option:`--outpu
 
    When ``processes`` is larger than 1 the pipeline may use the Python `multiprocessing` module to parallelize processing of multiple datasets across multiple processors.
    
-   .. note::
-
-      This option is provided for forward-compatibility, but is not yet supported by ``ap_verify``.
-
 .. option:: --metrics-file <filename>
 
    **Output metrics file.**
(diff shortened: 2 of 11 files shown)


bin/run_ci_dataset.sh
```diff
@@ -24,9 +24,8 @@
 # Simple script for running an entire dataset through ap_verify
 # Assumes that the requested dataset is already set up in EUPS
 #
-# The ap_verify workspace is a directory with the dataset name; until DM-13887
-# gets resolved, metrics are dumped to the workspace directory with names
-# containing each run's data ID.
+# The ap_verify workspace is a directory with the dataset name; metrics are
+# dumped to the workspace directory with names containing each run's data ID.
```
mrawls: Would it be possible to dump metrics to the output directory so they aren't overwritten if the same dataId is reprocessed later?

doc/lsst.ap.verify/command-line-reference.rst
```diff
@@ -36,20 +36,16 @@ If the pipeline fails, the status code will be an interpreter-dependent nonzero
...
 
 .. option:: --id <dataId>
 
    **Butler data ID.**
 
-   The input data ID is required for all ``ap_verify`` runs except when using :option:`--help`.
-
    Specify data ID to process using data ID syntax.
    For example, ``--id "visit=12345 ccd=1 filter=g"``.
+   If this argument is omitted, then all data IDs in the dataset will be processed.
```
mrawls: Good call!

doc/lsst.ap.verify/command-line-reference.rst
```diff
@@ -81,10 +77,9 @@ Required arguments are :option:`--dataset`, :option:`--id`, and :option:`--outpu
 
    **Output metrics file.**
 
-   The name of a file to contain the metrics measured by ``ap_verify``, in a format readable by the :doc:`lsst.verify</modules/lsst.verify/index>` framework.
-   If omitted, the output will go to a file named :file:`ap_verify.verify.json` in the user's working directory.
-
-   This argument can be used to run multiple instances of ``ap_verify`` concurrently, with each instance producing output to a different metrics file.
+   The template for a file to contain metrics measured by ``ap_verify``, in a format readable by the :doc:`lsst.verify</modules/lsst.verify/index>` framework.
+   The string ``{dataId}`` shall be replaced with the data ID associated with the job, and its use is strongly recommended.
+   If omitted, the output will go to files named after ``ap_verify.{dataId}.verify.json`` in the user's working directory.
```
mrawls: Oh... haha... still, I think a putting these in the "output" directory might be a better default.

python/lsst/ap/verify/metrics.py
```diff
@@ -76,16 +76,31 @@ class MetricsParser(argparse.ArgumentParser):
...
                           squashUrl=os.getenv(_ENV_URL, _SQUASH_DEFAULT_URL))
 
 
+# borrowed from validate_drp
+def _extract_instrument_from_butler(butler):
+    """Extract the last part of the mapper name from a Butler repo.
+    'lsst.obs.lsstSim.lsstSimMapper.LsstSimMapper' -> 'LSSTSIM'
+    'lsst.obs.cfht.megacamMapper.MegacamMapper' -> 'CFHT'
+    'lsst.obs.decam.decamMapper.DecamMapper' -> 'DECAM'
+    'lsst.obs.hsc.hscMapper.HscMapper' -> 'HSC'
```
mrawls: Might be worth pointing out these are examples and it will still return an uppercase instrument name even if it's not in this list (for way in the future when our code works with every telescope ever)

python/lsst/ap/verify/metrics.py
```diff
@@ -98,15 +113,28 @@ class AutoJob:
 
     Parameters
     ----------
+    butler : `lsst.daf.persistence.Butler`
+        the repository associated with this ``Job``.
```
mrawls: I feel like you've slipped into starting your docstrings with lower case somewhere around here, and beginning each "sentence" with an uppercase letter would be more in line with our style guide.

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -66,98 +46,50 @@ class ApPipeParser(argparse.ArgumentParser):
...
+    args = [workspace.dataRepo,
+            "--output", workspace.outputRepo,
+            "--calib", workspace.calibRepo,
+            "--template", workspace.templateRepo]
+    args.extend(_getConfigArguments(workspace))
+    if parsedCmdLine.dataId:
+        args.extend(["--id", *parsedCmdLine.dataId.split(" ")])
+    else:
+        args.extend(["--id"])
+    args.extend(["--processes", str(parsedCmdLine.processes)])
```
mrawls: Hooray for forward-thinking optimism 💯

tests/test_driver.py
```diff
@@ -120,99 +109,49 @@ def setUpMockPatch(self, target, **kwargs):
...
-
-        self.assertEqual(self.job.measurements, self.subtaskJob.measurements)
-
-    # Mock up ApPipeTask to avoid doing any processing.
-    @unittest.mock.patch("lsst.ap.verify.pipeline_driver._getConfig", return_value=None)
     @patchApPipe
-    def testUpdateMetricsOnError(self, _mockConfig, mockClass):
-        """Test that runApPipe stores metrics in a job even when the pipeline fails.
+    def testRunApPipeDataIdReporting(self, mockClass):
+        """Test that runApPipe reports the data IDs that were processed.
```
mrawls: Thank you for testing this!

python/lsst/ap/verify/pipeline_driver.py
kfindeisen: It supports multiprocessing as of this ticket...? 😕

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -66,98 +46,50 @@ class ApPipeParser(argparse.ArgumentParser):
...
         # Help and documentation will be handled by main program's parser
         argparse.ArgumentParser.__init__(self, add_help=False)
-        self.add_argument('--id', dest='dataId', required=True,
-                          help='An identifier for the data to process. '
-                          'May not support all features of a Butler dataId; '
-                          'see the ap_pipe documentation for details.')
+        self.add_argument('--id', dest='dataId', default="",
+                          help='An identifier for the data to process.')
         self.add_argument("-j", "--processes", default=1, type=int,
                           help="Number of processes to use. Not yet implemented.")
```
mrawls: This is what I was basing my "optimism" comment on; your statement here that multiprocessing (i.e., with the -j keyword) is not yet implemented. If it is implemented now that's even better, and please update this docstring!

kfindeisen: I could swear I fixed that. Good catch, thanks!

doc/lsst.ap.verify/command-line-reference.rst
kfindeisen: Good idea, but I'd rather do this on a new ticket, since it's a change from the pre-existing behavior and I'd want to double-check that custom filenames don't break things (including CI!) in surprising ways.

PR #59, opened 2019-01-18
DM-16535: Implement MetricRegistry

kfindeisen:
This PR adds the configuration information required by `MetricsControllerTask` as of lsst/verify#33.

### DIFF
--- a/config/default_metrics.py
+++ b/config/default_metrics.py
@@ -1,0 +1,27 @@
+from lsst.ap.verify.measurements.profiling import TimingMetricConfig
+
+config.jobFileTemplate = "ap_verify.metricTask{id}.{dataId}.verify.json"
+
+config.measurers = ["timing"]
+
+timingConfigs = {
+    "apPipe.runDataRef": "ap_pipe.ApPipeTime",
+    "apPipe:ccdProcessor.runDataRef": "pipe_tasks.ProcessCcdTime",
+    "apPipe:ccdProcessor:isr.runDataRef": "ip_isr.IsrTime",
+    "apPipe:ccdProcessor:charImage.runDataRef": "pipe_tasks.CharacterizeImageTime",
+    "apPipe:ccdProcessor:calibrate.runDataRef": "pipe_tasks.CalibrateTime",
+    "apPipe:differencer.runDataRef": "pipe_tasks.ImageDifferenceTime",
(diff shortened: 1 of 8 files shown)


python/lsst/ap/verify/ap_verify.py
```diff
@@ -143,9 +147,13 @@ def _measureFinalProperties(workspace, dataIds, args):
     args : `argparse.Namespace`
         Command-line arguments, including arguments controlling output.
     """
+    metricFile = args.metrics_config \
+        if args.metrics_config is not None \
+        else os.path.join(lsst.utils.getPackageDir("ap_verify"), "config", "default_metrics.py")
```
parejkoj: I think in this case, a standard `if:else:` is clearer.

config/default_metrics.py
```diff
@@ -0,0 +1,27 @@
...
+    "apPipe:ccdProcessor:isr.runDataRef": "ip_isr.IsrTime",
+    "apPipe:ccdProcessor:charImage.runDataRef": "pipe_tasks.CharacterizeImageTime",
+    "apPipe:ccdProcessor:calibrate.runDataRef": "pipe_tasks.CalibrateTime",
+    "apPipe:differencer.runDataRef": "pipe_tasks.ImageDifferenceTime",
+    "apPipe:differencer:astrometer.loadAndMatch": "meas_astrom.AstrometryTime",
+    "apPipe:differencer:register.run": "pipe_tasks.RegisterImageTime",
+    "apPipe:differencer:subtract.subtractExposures": "ip_diffim.ImagePsfMatchTime",
+    "apPipe:differencer:detection.run": "meas_algorithms.SourceDetectionTime",
+    "apPipe:differencer:measurement.run": "ip_diffim.DipoleFitTime",
+    "apPipe:associator.run": "ap_association.AssociationTime",
```
parejkoj: It seems unfortunate that the names had to be lifted out of the yaml and stringified: makes it a bit harder to scan. But probably inevitable.

kfindeisen: I think this is a usability improvement overall, since "what tasks do we want to run?" is something Stack users expect to live in a Python config file. The only real problem is that for verification we want to run a *lot* of tasks.

I've discussed the "lots of tasks" aspect with @TallJimbo and he said the Gen 3 framework will eventually support grouping pipeline tasks in some way. Hopefully that will, if not make config files simpler, at least let us segregate the parts that are and are not likely to change.

PR #63, opened 2019-03-01
DM-16536: Migrate all metrics from ap.verify.measurements

kfindeisen:
This PR replaces all functions previously in `lsst.ap.verify.measurements` with `MetricTask` equivalents in other packages (except for `TimingMetricTask`, which cannot be reasonably placed in any other Stack package). The hardcoded references inside `measurements/compute_metrics.py` are replaced with `MetricTaskConfig` instances; two default configs for supporting CI are provided in `config/`.

Correctly handling `lsst.ap.association.TotalUnassociatedDiaObjectsMetricTask` required a significant rewrite of the post-`ApPipeTask` control flow, which had been written on the assumption that all metrics are calculated after every image. The current solution (hardcoding the existence of two `MetricsControllerTask`s, one for image- and one for dataset-level metrics) is clumsy both for the user and for code maintainers, but I believe it is better to defer generalizing the code to another, more focused ticket.


kfindeisen:
It came up in the [RFC-550 discussion](https://jira.lsstcorp.org/browse/RFC-550?focusedCommentId=181136#comment-181136); basically there is no other place to put it at the moment. Where would be a good place to document a decision like that?

jonathansick:
Oh I see. I don't entirely understand how that circular dependency comes up. I'll have to take a closer look at how these tasks work. Until I understand the nature of the issue I can't really say if or where it should be documented.

kfindeisen:
`MetricTask` (Gen 2 or Gen 3, it doesn't matter) inherits from `lsst.pipe.base.Task`, so `pipe_base` must be built before `verify`. A hypothetical `lsst.pipe.base.TimingMetricTask` would inherit from `MetricTask`, so `verify` would need to be built before `pipe_base`.

jonathansick:
Sorry for being confused, but I thought the hope would be to migrate `TimingMetricTask` to `lsst.verify`, not `lsst.pipe.base`.

kfindeisen:
Ah. Yes, that would work from a dependency point of view, but I thought the RFC consensus was that code associated with specific metrics would not go into `lsst.verify`. We could make an exception, but that makes me wonder if there would be other exceptions later.

kfindeisen:
@jonathansick so should I take any action on `TimingMetricTask`? Everything else is ready to go.

jonathansick:
By all means, this is good to go.

> We could make an exception, but that makes me wonder if there would be other exceptions later.

So I definitely feel that something like AM1MetricTask, or some other very domain-specific measurement task, ought to belong in the package most associated the metric. But something like timing (if I understand the task's intent well enough) seems so broadly applicable that it could belong in `lsst.verify`. Do you agree with that?

kfindeisen:
@jonathansick Fine. Any preferences on whether `profiling.py` goes into a specific subpackage or directly into `lsst.verify`? I'm a bit worried about namespace clutter/confusion, but it's your package.

kfindeisen:
Given the issue I just ran into with documentation builds, I'll merge as-is and open another ticket for moving `TimingMetricTask` to `verify`.

PR #64, opened 2019-03-13
DM-16606: ap_pipe should not create DB automatically

kfindeisen:
This PR gives the pipeline driver a call to the `make_ppdb.py` script added by lsst/ap_pipe#39, and updates unit tests to require that it be called.

### DIFF
--- a/python/lsst/ap/verify/pipeline_driver.py
+++ b/python/lsst/ap/verify/pipeline_driver.py
@@ -34,6 +34,7 @@ import os
 
 import lsst.log
 import lsst.ap.pipe as apPipe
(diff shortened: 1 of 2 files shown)


python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -34,6 +34,7 @@
 
 import lsst.log
 import lsst.ap.pipe as apPipe
+import lsst.ap.pipe.make_ppdb as makePpdb
```
andy-slac: Do you really need to rename it? Whey not `from lsst.ap.pipe.make_ppdb import makePpdb` or `from lsst.ap.pipe import make_ppdb`?

PR #66, opened 2019-03-19
DM-17248: ap_verify should create .json files in the workspace by default

kfindeisen:
This PR modifies the default location of `ap_verify`'s output `Job` files to the workspace created by `--output`. This change is backwards-compatible with scripts that explicitly assign a location.

### DIFF
--- a/python/lsst/ap/verify/metrics.py
+++ b/python/lsst/ap/verify/metrics.py
@@ -31,6 +31,7 @@ processing of individual measurements. Measurements are handled in the
 __all__ = ["MetricsParser", "computeMetrics"]
 
 import argparse
+import collections
 import copy
 import os
 import warnings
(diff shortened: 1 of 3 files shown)


python/lsst/ap/verify/metrics.py
```diff
@@ -91,6 +93,15 @@ def __call__(self, _parser, _namespace, _values, option_string=None):
         warnings.warn(message, category=FutureWarning)
 
 
+class _OptionalFormatDict(collections.UserDict):
+    """A dictionary that, when used with a formatter, preserves unknown keys.
```
mrawls: I'm not understanding why this OptionalFormatDict is necessary here, can you please explain?

tests/test_workspace.py
```diff
@@ -49,7 +49,7 @@ def _assertInDir(self, path, baseDir):
         self.assertEqual(ancestor, _canonDir)
 
     def _assertNotInDir(self, path, baseDir):
-        """Test that ``path`` is a subpath of ``baseDir``.
+        """Test that ``path`` is not a subpath of ``baseDir``.
```
mrawls: Good catch!

python/lsst/ap/verify/metrics.py
kfindeisen: Most of the string substitution (in particular, `{dataId}`) is done by `lsst.verify.gen2tasks.MetricsControllerTask`, so `ap_verify` shouldn't mess with it -- it should only substitute `{output}`, which is specific to `ap_verify`. But `str.format` raises an exception if it finds a replacement field it doesn't know how to fill in. So this class is basically a hack :pick: that makes `str.format_map` think it's replacing all the fields, but the replacements other than `{output}` are no-ops.

I'm not sure if that answers your question.

mrawls: That at least makes sense. It still appears convoluted to me, but I don't have an alternative to suggest. Can you please add a sentence or two to the docstring which explains the existence of this class (similar to what you commented here)?

PR #68, opened 2019-05-31
DM-19857: Update for new defects ingestion.

SimonKrughoff:
In light of DM-18739, we needed to update how this package manages that.

### DIFF
--- a/python/lsst/ap/verify/dataset.py
+++ b/python/lsst/ap/verify/dataset.py
@@ -139,12 +139,6 @@ class Dataset:
         return os.path.join(self.datasetRoot, 'calib')
 
     @property
-    def defectLocation(self):
-        """The directory containing defect files (`str`, read-only).
-        """
-        return os.path.join(getPackageDir('obs_decam_data'), 'decam', 'defects')
-
-    @property
     def refcatsLocation(self):
         """The directory containing external astrometric and photometric
         reference catalogs (`str`, read-only).
@@ -200,8 +194,6 @@ class Dataset:
             raise RuntimeError('Dataset at ' + self.datasetRoot + 'is missing data directory')
         if not os.path.exists(self.calibLocation):
             raise RuntimeError('Dataset at ' + self.datasetRoot + 'is missing calibration directory')
-        if not os.path.exists(self.defectLocation):
-            raise RuntimeError('Dataset at ' + self.datasetRoot + 'is missing defect directory')
         # Template and refcat directories might not be subdirectories of self.datasetRoot
         if not os.path.exists(self.templateLocation):
             raise RuntimeError('Dataset is missing template directory at ' + self.templateLocation)
--- a/python/lsst/ap/verify/dataset.py
+++ b/python/lsst/ap/verify/dataset.py
@@ -204,7 +204,7 @@ class Dataset:
         if not _isRepo(self._stubInputRepo):
             raise RuntimeError('Stub repo at ' + self._stubInputRepo + 'is missing mapper file')
 
-    def makeCompatibleRepo(self, repoDir):
+    def makeCompatibleRepo(self, repoDir, calibRepoDir):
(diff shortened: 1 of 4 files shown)


python/lsst/ap/verify/dataset.py
```diff
@@ -142,7 +142,7 @@ def calibLocation(self):
     def defectLocation(self):
         """The directory containing defect files (`str`, read-only).
         """
-        return self.calibLocation
+        return os.path.join(getPackageDir('obs_decam_data'), 'decam', 'defects')
```
kfindeisen: Please do not hard-code dependencies on specific instruments into `ap_verify`. If defects are no longer part of the dataset format, then removing the `defectLocation` member would be more appropriate.

python/lsst/ap/verify/ingestion.py
```diff
@@ -303,15 +293,11 @@ def _ingestDefects(self, dataset, workspace):
...
-            if self.config.defectTarball:
-                self.log.info("Ingesting defects...")
-                defectFile = os.path.join(dataset.defectLocation, self.config.defectTarball)
-                self._doIngestDefects(workspace.dataRepo, workspace.calibRepo, defectFile)
-                self.log.info("Defects are now ingested in {0}".format(workspace.calibRepo))
-            else:
-                self.log.info("No defects to ingest, skipping...")
+            self.log.info("Ingesting defects...")
+            self._doIngestDefects(workspace.dataRepo, workspace.calibRepo, dataset.defectLocation)
+            self.log.info("Defects are now ingested in {0}".format(workspace.calibRepo))
```
kfindeisen: Does there need to be a way to disable defect ingestion for instruments that don't use it (e.g., HSC), or is that something `IngestDefectsTask` takes care of?

mrawls: It looks like the changes in obs_decam result in ingested defects always landing in obs_decam_data. Do we even need to deal with defects in ap_verify **at all** anymore?

python/lsst/ap/verify/dataset.py
SimonKrughoff: Removing defectLocation required more surgery than I had hoped to do.  Do you think you could have a look and suggest how best to do that?

python/lsst/ap/verify/ingestion.py
SimonKrughoff: They do not land in `obs_decam_data`.  They land in `workspace.calibRepo`.

They will need to be ingested in order to use them.

SimonKrughoff:
As I say above, they are not ingested into a calib repo in `obs_decam_data`.  They still need to be put in a calib repository to use them.

python/lsst/ap/verify/dataset.py
kfindeisen: It looks like, outside of test code, the only place where `defectLocation` gets used is in your call to `_doIngestDefects`. So it might make more sense to make that a config field in `DatasetIngestConfig`, which will at least be consistent with how the inputs to `IngestCalibsTask` are handled.

The defect target could be configured in the dataset-specific config files, but it could also be put in an obs package config override file (already supported as `datasetIngest.py`). I think you need to modify those files anyway because you removed config fields.

SimonKrughoff: @kfindeisen I've implemented your suggestions.  I've updated `obs_test` and `obs_decam` though on a different ticket since they have to be merged together anyway: tickets/DM-18739 and tickets/DM-19730 respectively.  I didn't see anything to change in `obs_subaru`.  I hope this addresses your concerns.

@mrawls if you are happy with things as they stand, will you approve this please, or let me know what else I need to do?

kfindeisen:
There's a reference to defect files in `doc/lsst.ap.verify/datasets-creation.rst`. Would it be possible to remove that before merging?

PR #69, opened 2019-08-02
DM-20570: Pipeline failure treated as ap_verify success

kfindeisen:
This PR propagates pipeline status information from the `ApPipeTask` wrapper to the main function, where it can be returned to the shell as if for a command-line task. This fixes a bug where `ap_verify.py` would return 0 after catastrophic pipeline failure and the Jenkins run would be marked as successful.

I think `ap_verify`'s exit code is correctly propagated from `run_ci_dataset.sh` up to Jenkins, but I'm not sure how to test that. :confused:

### DIFF
--- a/python/lsst/ap/verify/ap_verify.py
+++ b/python/lsst/ap/verify/ap_verify.py
@@ -141,6 +141,12 @@ def runApVerify(cmdLine=None):
     cmdLine : `list` of `str`
         an optional command line used to execute `runApVerify` from other
         Python code. If `None`, `sys.argv` will be used.
+
+    Returns
+    -------
+    nFailed : `int`
+        The number of data IDs that were not successfully processed, up to 127,
+        or 127 if the task runner framework failed.
     """
     lsst.log.configure()
     log = lsst.log.Log.getLogger('ap.verify.ap_verify.main')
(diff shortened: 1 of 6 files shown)


python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -65,8 +65,9 @@ def runApPipe(workspace, parsedCmdLine):
 
     Returns
     -------
-    dataIds : `lsst.pipe.base.DataIdContainer`
-        The set of complete data IDs fed into ``ap_pipe``.
+    result : `Struct`
+        The `Struct` returned from `ApPipeTask.parseAndRun` with
+        ``doReturnResults=False``.
```
mrawls: This is the part where I express dismay and/or confusion that we return `result` which explicitly contains something that says `doReturnResults=False`. Can you give me a little more context for the confusing terminology here?

python/lsst/ap/verify/ap_verify.py
```diff
@@ -152,8 +157,11 @@ def runApVerify(cmdLine=None):
     ingestDataset(args.dataset, workspace)
 
     log.info('Running pipeline...')
-    expandedDataIds = runApPipe(workspace, args)
-    computeMetrics(workspace, expandedDataIds, args)
+    apPipeResults = runApPipe(workspace, args)
+    computeMetrics(workspace, apPipeResults.parsedCmd.id, args)
+
+    # ApPipeTaskRunner does not override default results handling, exitStatus always defined
+    return sum(((res.exitStatus != 0) for res in apPipeResults.resultList))
```
kfindeisen: This code is unsafe, because lsst/afw#97 allows `resultList` to be `None` (not clear if its elements can be `None` or not).

python/lsst/ap/verify/pipeline_driver.py
kfindeisen: `doReturnResults` is a [keyword argument to `parseAndRun`](https://pipelines.lsst.io/py-api/lsst.pipe.base.CmdLineTask.html#lsst.pipe.base.CmdLineTask.parseAndRun). I think the logic is that `parseAndRun` returns things that are not results? 🤷‍♂

python/lsst/ap/verify/ap_verify.py
mrawls: Hmm, oh dear, good real-time catch. Could you fix this with a try/except for the return value and have it return something else if summing the exitStatuses doesn't work?

PR #77, opened 2019-11-21
DM-21155: Switch to new ingest name

SimonKrughoff:


python/lsst/ap/verify/ingestion.py
```diff
@@ -100,7 +100,7 @@ class DatasetIngestConfig(pexConfig.Config):
         doc="Path to top level of the defect tree.  This is a directory with a directory per sensor."
     )
     defectIngester = pexConfig.ConfigurableField(
-        target=IngestDefectsTask,
+        target=IngestStdTextCalibsTask,
```
kfindeisen: Is `IngestStdTextCalibsTask` a drop-in replacement for `IngestDefectsTask`? If there are any API changes, our current tests won't pick them up, because this code is no longer used for DECam data.

kfindeisen: Looking at lsst/obs_decam#123, the command-line interface is the same, and that's unfortunately what `DatasetIngestTask` has to use.

SimonKrughoff: Are there more changes required?  It was meant as a drop-in replacement.

kfindeisen: No, there aren't.

PR #86, opened 2020-05-07
DM-24324: Generate alert packets in the alert production pipeline

morriscb:


python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -132,5 +132,7 @@ def _getConfigArguments(workspace):
     # ApVerify will use the sqlite hooks for the Apdb.
     args.extend(["--config", "diaPipe.apdb.db_url=sqlite:///" + workspace.dbLocation])
     args.extend(["--config", "diaPipe.apdb.isolation_level=READ_UNCOMMITTED"])
+    # Put output alerts into the workspace.
+    args.extend(["--config", "diaPipe.alertPackager.alertWriteLocation=" + workspace.outputRepo + "/alerts"])
```
kfindeisen, line 136: Not sure I like having this command-line config override (see my comment [in ap_association](https://github.com/lsst/ap_association/pull/79#discussion_r422381313)).

If the alert output is here to stay and it's not supposed to be part of the repository, please consider handling it the same way as `workspace.dbLocation` instead of putting it in the repository (if nothing else, it will be more compatible with Gen 3!)

PR #90, opened 2020-06-03
DM-25216: ap_verify failure: ap_verify.py: error: no config field: diaPipe.doSerializeAlerts

jdswinbank:


morriscb:
Thanks for catching this

PR #91, opened 2020-06-12
DM-25407: ap_verify cannot handle curated crosstalk data in Gen 2

kfindeisen:
This PR does a bulk removal of all references to "defects" in `DatasetIngestTask`, removing some out-of-date documentation in the process. It then modifies the code to take a (possibly empty) list of curated calib directories, allowing more than one type of curated calib to be ingested at a time.

I ended up sacrificing the code that guards against repeated ingestion of defects. It should be possible to make the check more generic (auto-detecting what the ingested files should look like), but I'm putting that off until [DM-25414](https://jira.lsstcorp.org/browse/DM-25414) to get things merged today.

### DIFF
--- a/python/lsst/ap/verify/ingestion.py
+++ b/python/lsst/ap/verify/ingestion.py
@@ -93,16 +93,16 @@ class DatasetIngestConfig(pexConfig.Config):
         default=9999,
         doc="Calibration validity period (days). Assumed equal for all calib types.")
 
-    textDefectPath = pexConfig.Field(
+    curatedCalibPaths = pexConfig.Field(
         dtype=str,
         default=None,
         optional=True,
         doc="Path to top level of the defect tree.  This is a directory with a directory per sensor. "
             "Set to None to disable defect ingestion."
     )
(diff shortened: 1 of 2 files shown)


python/lsst/ap/verify/ingestion.py
```diff
@@ -93,16 +93,15 @@ class DatasetIngestConfig(pexConfig.Config):
...
-    textDefectPath = pexConfig.Field(
+    curatedCalibPaths = pexConfig.ListField(
         dtype=str,
-        default=None,
-        optional=True,
-        doc="Path to top level of the defect tree.  This is a directory with a directory per sensor. "
-            "Set to None to disable defect ingestion."
+        default=[],
+        doc="Paths to top level of each curated calib's tree. Each is a "
+            "directory with a directory per sensor."
```
mrawls, line 100: Rephrase: "Paths to the top level of each curated calib's tree (e.g., defects, crosstalk). Each path should be a directory which contains one subdirectory per sensor."

PR #92, opened 2020-07-07
DM-20922: ap_verify can't handle --id with empty argument

kfindeisen:
This PR lets callers of `ap_verify.py` use `--id` by itself, with no data ID, without changing the existing behavior when omitting `--id` or using it with a string argument.

### DIFF
--- a/python/lsst/ap/verify/pipeline_driver.py
+++ b/python/lsst/ap/verify/pipeline_driver.py
@@ -48,13 +48,29 @@ class ApPipeParser(argparse.ArgumentParser):
     def __init__(self):
         # Help and documentation will be handled by main program's parser
         argparse.ArgumentParser.__init__(self, add_help=False)
-        self.add_argument('--id', dest='dataIds', action='append', default=[],
+        # namespace.dataIds will always be a list of 0 or more nonempty strings, regardless of inputs.
+        # TODO: in Python 3.8+, action='extend' handles nargs='?' more naturally than 'append'.
+        self.add_argument('--id', dest='dataIds', action=self.AppendOptional, nargs='?', default=[],
                           help='An identifier for the data to process.')
         self.add_argument("-j", "--processes", default=1, type=int,
                           help="Number of processes to use.")
         self.add_argument("--skip-pipeline", action="store_true",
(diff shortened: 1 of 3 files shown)


python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -48,14 +48,30 @@ class ApPipeParser(argparse.ArgumentParser):
...
         self.add_argument("--skip-pipeline", action="store_true",
                           help="Do not run the AP pipeline itself. This argument is useful "
                                "for testing metrics on a fixed data set.")
 
+    class AppendOptional(argparse.Action):
+        """A variant of the built-in "append" action that ignores None values
+        instead of appending them.
+        """
+        # This class can't inherit from the built-in action because the
+        # corresponding class is implementation-dependent
```
parejkoj, line 66: Period after sentence.

Also, it took me a couple readings to fully parse this. I think you mean the implementation of `action='append'` can vary?

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -48,14 +48,30 @@ class ApPipeParser(argparse.ArgumentParser):
     def __init__(self):
         # Help and documentation will be handled by main program's parser
         argparse.ArgumentParser.__init__(self, add_help=False)
-        self.add_argument('--id', dest='dataIds', action='append', default=[],
+        # namespace.dataIds will always be a list of 0 or more nonempty strings, regardless of inputs.
+        # TODO: in Python 3.8+, action='extend' handles nargs='?' more naturally than 'append'.
```
parejkoj, line 52: Useful note. Probably not worth filing a separate ticket for it, though?

python/lsst/ap/verify/pipeline_driver.py
kfindeisen: I mean that there's no official class that implements `action='append'`, so presumably different Python implementations can call it whatever they like. In the version provided by conda-forge, it's `argparse._AppendAction`.

parejkoj: Ah, maybe include that in the comment then "no public class implements this action". That's a lot clearer to me.

PR #99, opened 2020-08-13
DM-24262: Run HSC AP processing in CI using Gen 3

kfindeisen:
This PR adds a machine-friendly flag to `run_ci_dataset.sh` that can be passed to `ap_verify.py`.

### DIFF
--- a/bin/run_ci_dataset.sh
+++ b/bin/run_ci_dataset.sh
@@ -35,17 +35,19 @@ print_error() {
 
 usage() {
     print_error
-    print_error "Usage: $0 -d DATASET [-h]"
+    print_error "Usage: $0 -d DATASET [-g NUM] [-h]"
     print_error
     print_error "Specifc options:"
     print_error "   -d          Dataset name"
+    print_error "   -g          Middleware generation number (int)"
     print_error "   -h          show this message"
     exit 1
 }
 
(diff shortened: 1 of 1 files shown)


bin/run_ci_dataset.sh
```diff
@@ -35,17 +35,19 @@ print_error() {
 
 usage() {
     print_error
-    print_error "Usage: $0 -d DATASET [-h]"
+    print_error "Usage: $0 -d DATASET [-g NUM] [-h]"
     print_error
     print_error "Specifc options:"
     print_error "   -d          Dataset name"
+    print_error "   -g          Generation number (int)"
```
ktlim, line 42: Maybe "Butler" or "Middleware" generation number would be clearer if more verbose?

bin/run_ci_dataset.sh
```diff
@@ -35,17 +35,19 @@ print_error() {
 
 usage() {
     print_error
-    print_error "Usage: $0 -d DATASET [-h]"
+    print_error "Usage: $0 -d DATASET [-g NUM] [-h]"
     print_error
     print_error "Specifc options:"
```
jdswinbank, line 40: (Just noticed this going by in my mail and though “hey, maybe this'd be a good opportunity to fix that typo...”)

PR #101, opened 2020-08-27
DM-26237: Can't rerun ap_verify on same repository in Gen 3

kfindeisen:
This PR fixes a bug in how the `ap_verify` dataset is turned into a workspace repository in Gen 3, and fixes the test case that was supposed to enforce the desired behavior.

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -115,6 +115,20 @@ class DatasetTestSuite(DataTestCase):
             if os.path.exists(testDir):
                 shutil.rmtree(testDir, ignore_errors=True)
 
+    def _checkOutputGen3(self, repo):
+        """Perform various integrity checks on a repository.
+
+        Parameters
+        ----------
+        repo : `str`
+            The repository to test. Currently only filesystem repositories
+            are supported.
+        """
+        self.assertTrue(os.path.exists(repo), 'Output directory must exist.')
+        # Call to Butler will fail if repo is corrupted
+        butler = dafButler.Butler(repo)
+        self.assertIn("LSST-ImSim/calib", butler.registry.queryCollections())
+
     def testOutputGen3(self):
         """Verify that a Dataset can create an output repository as desired.
         """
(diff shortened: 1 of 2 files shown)


tests/test_dataset.py
```diff
@@ -115,6 +115,14 @@ def testExistingOutput(self):
             if os.path.exists(testDir):
                 shutil.rmtree(testDir, ignore_errors=True)
 
+    def _checkOutputGen3(self, repo):
+        self.assertTrue(os.path.exists(repo), 'Output directory must exist.')
+        self.assertTrue(os.listdir(repo), 'Output directory must not be empty.')
+        self.assertTrue(os.path.exists(os.path.join(repo, 'butler.yaml')),
+                        'Output directory must have a butler.yaml file.')
+        butler = dafButler.Butler(repo)
```
timj, line 123: Won't this do the first three tests implicitly in that the Butler constructor would fail? Maybe comment here that you are being explicit in testing because you want to give a more compact reason for the problem?

Also note in a docstring for the method that the tests all require a local file Butler repo and won't work for the general case. (making this work generally is straightforward: use ButlerURI -- not worth it for now though)

tests/test_dataset.py
```diff
@@ -141,18 +144,10 @@ def testExistingOutputGen3(self):
...
             self._testbed.makeCompatibleRepoGen3(outputDir)
-            self.assertTrue(os.path.exists(outputDir), 'Output directory must exist.')
-            self.assertTrue(os.listdir(outputDir), 'Output directory must not be empty.')
-            self.assertTrue(os.path.exists(os.path.join(outputDir, 'butler.yaml')),
-                            'Output directory must have a butler.yaml file.')
-            butler = dafButler.Butler(outputDir)
-            self.assertIn("LSST-ImSim/calib", butler.registry.queryCollections())
+            self._checkOutputGen3(outputDir)
+            self._testbed.makeCompatibleRepoGen3(outputDir)
+            self._checkOutputGen3(outputDir)
```
timj, line 150: This second check is making sure that the previous line didn't delete it?

kfindeisen: Making sure nothing's been invalidated. I don't see any point in filtering out tests on the grounds that "surely that couldn't break".

tests/test_dataset.py
kfindeisen: Does that really call for a comment? I think it's fairly standard practice to have assertions that test preconditions for later assertions, and some testing libraries provide explicit support for this (`unittest` doesn't).

I'll document that the code only supports directory-based repositories at the moment.

timj: My main thought is that two of the tests could be construed as internal implementation details for butler (files being present and there being a file called butler.yaml). All that really matters is that you can instantiate a butler with that root.

PR #102, opened 2020-09-11
DM-25806: Support parallel ap_verify ingestion in Gen 3

kfindeisen:
This PR implements multiprocessing in `Gen3DatasetIngestTask` after adding basic value support to the `Dataset` and `Workspace` classes to make them more unit-test-friendly. This required some refactoring of `ap_verify`'s command-line argument parser.

### DIFF
--- a/tests/test_ingestion.py
+++ b/tests/test_ingestion.py
@@ -22,6 +22,7 @@
 #
 
 import os
+import pickle
 import shutil
 import tempfile
 import unittest.mock
@@ -470,6 +471,19 @@ class IngestionTestSuiteGen3(DataTestCase):
             ingestion._findMatchingFiles(testDir, ['*.fits.gz'], exclude=['calib']),
             allFiles
         )
+
+    def testPickling(self):
+        """Test that a Gen3DatasetIngestTask can be pickled correctly.
+
+        This is needed for multiprocessing support.
+        """
+        stream = pickle.dumps(self.task)
+        copy = pickle.loads(stream)
+        self.assertEqual(self.task.getFullName(), copy.getFullName())
+        self.assertEqual(self.task.log.getName(), copy.log.getName())
+        # Equality for config ill-behaved; skip testing it
+        self.assertEqual(self.task.dataset, copy.dataset)
+        self.assertEqual(self.task.workspace, copy.workspace)
 
 
 class MemoryTester(lsst.utils.tests.MemoryTestCase):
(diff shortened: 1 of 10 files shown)


tests/test_ingestion.py
```diff
@@ -471,6 +472,19 @@ def testFindMatchingFiles(self):
...
+    def testPickling(self):
+        """Test that a Gen3DatasetIngestTask can be pickled correctly.
+
+        This is needed for multiprocessing support.
+        """
+        stream = pickle.dumps(self.task)
+        copy = pickle.loads(stream)
+        self.assertEqual(self.task.getFullName(), copy.getFullName())
+        self.assertEqual(self.task.log.getName(), copy.log.getName())
+        # Equality for config ill-behaved; skip it
```
isullivan, line 484: This is a minor nitpick, but could you rephrase this comment? "Skip it" suggests to me that code including a test for equality of the config is present and is skipped, while what is happening here is that code doesn't exist (for a good reason). Maybe something like "Do not test equality for the config because it is ill-behaved"?

tests/test_workspace.py
```diff
@@ -35,12 +35,29 @@
 class WorkspaceGen2TestSuite(lsst.utils.tests.TestCase):
 
     def setUp(self):
-        self._testWorkspace = tempfile.mkdtemp()
+        self._testWorkspace = os.path.realpath(tempfile.mkdtemp())  # Avoid link problems
```
isullivan, line 38: Could you expand the comment to say "Use os.path.realpath to avoid link problems"?

tests/test_workspace.py
```diff
@@ -116,12 +133,29 @@ def testAlerts(self):
 class WorkspaceGen3TestSuite(lsst.utils.tests.TestCase):
 
     def setUp(self):
-        self._testWorkspace = tempfile.mkdtemp()
+        self._testWorkspace = os.path.realpath(tempfile.mkdtemp())  # Avoid link problems
```
isullivan, line 136: Please expand this comment as above.

isullivan (approved):
Looks great.

tests/test_ingestion.py
kfindeisen: IIRC `Config.__eq__` does exist, but it doesn't handle `ConfigurableField` very well. I probably should have filed a ticket, but was too distracted at the time.

PR #106, opened 2020-10-01
DM-26953: Update ap_verify system for calibration collections

kfindeisen:
This PR makes `ap_verify`'s Butler interface more specific by replacing general "all collections" queries with a list of expected collections. Recent changes to calibration handling have added collections that should not be used for user input.

### DIFF
--- a/python/lsst/ap/verify/pipeline_driver.py
+++ b/python/lsst/ap/verify/pipeline_driver.py
@@ -31,11 +31,12 @@ __all__ = ["ApPipeParser", "runApPipeGen2", "runApPipeGen3"]
 
 import argparse
 import os
+import re
 
 import lsst.log
 from lsst.utils import getPackageDir
-import lsst.daf.butler as dafButler
 import lsst.pipe.base as pipeBase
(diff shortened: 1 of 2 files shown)


python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -257,8 +258,14 @@ def _getCollectionArguments(workspace):
         following the conventions of `sys.argv`.
     """
     butler = workspace.workButler
-    inputs = set(butler.registry.queryCollections(collectionTypes={dafButler.CollectionType.RUN}))
-    inputs.discard(workspace.runName)
+    # Hard-code the repository names because it's hard to infer the inputs from the Butler
```
TallJimbo, line 261: Nomenclature in Gen3 is "collection names" rather than "repository names".

TallJimbo (approved):

python/lsst/ap/verify/pipeline_driver.py
kfindeisen: I have no idea why I said that... 😳

PR #107, opened 2020-10-06
DM-26662: ap_verify import slow in Gen 3

kfindeisen:
This PR creates the `ap_verify` repository without checksums, to ensure that copying the repository is a fast operation (minutes vs. hours, for large repositories). Testing suggests that `daf.butler.Config` silently ignores nonexistent keys, so the new code will work for both file-based and non-file datastores. However, this behavior doesn't seem to be documented anywhere.

### DIFF
--- a/python/lsst/ap/verify/dataset.py
+++ b/python/lsst/ap/verify/dataset.py
@@ -277,7 +277,10 @@ class Dataset:
         """
         # No way to tell makeRepo "create only what's missing"
         try:
-            repoConfig = dafButler.Butler.makeRepo(repoDir)
+            seedConfig = dafButler.Config()
+            # Checksums greatly slow importing of large repositories
+            seedConfig["datastore", "checksum"] = False
+            repoConfig = dafButler.Butler.makeRepo(repoDir, config=seedConfig)
             butler = dafButler.Butler(repoConfig, writeable=True)
             butler.import_(directory=self._preloadedRepo, filename=self._preloadedExport,
                            transfer="auto")


python/lsst/ap/verify/dataset.py
```diff
@@ -277,7 +277,10 @@ def makeCompatibleRepoGen3(self, repoDir):
         """
         # No way to tell makeRepo "create only what's missing"
         try:
-            repoConfig = dafButler.Butler.makeRepo(repoDir)
+            initConfig = dafButler.Config()
```
timj, line 280: This doesn't matter but in `butler create` command line we call this a seed config.

python/lsst/ap/verify/dataset.py
```diff
@@ -277,7 +277,10 @@ def makeCompatibleRepoGen3(self, repoDir):
...
+            # Checksums greatly slow importing of large repositories
+            initConfig[".datastore.checksum"] = False
```
timj, line 282: Inside butler we prefer to write this as `initConfig["datastore", "checksum"]` rather than relying on the delimiter at all. You can of course keep it with the leading delimiter.

timj (approved):
Looks fine. Regarding config validation, `Config` is just a clever dict. It knows nothing about the content. Later on this will be part of a `DatastoreConfig` that can do validation but will also ignore things it doesn't care about. These are not pex_config. That can be good and it can be bad. In this case it's good because if you end up with an in-memory datastore nothing cares about the weird extra.

python/lsst/ap/verify/dataset.py
kfindeisen: If you prefer that format, consider emphasizing it in the class docs. Right now they focus almost entirely on delimiters, and all the alternatives are presented as if they were equal.

PR #113, opened 2020-11-19
DM-26375: Allow databases other than SQLite when running ap_verify

kfindeisen:
This PR lets users override `db_url` when calling `ap_verify`. I've tested the changes with `lsst-pg-devel1` as well as custom SQLite stores.


isullivan (approved):
Looks good to me.

PR #114, opened 2020-11-25
DM-27147: Revert to older skymap dataset type names in processing.

TallJimbo:


python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -169,6 +169,14 @@ def runApPipeGen3(workspace, parsedCmdLine, processes=1):
     pipelineArgs.extend(["--processes", str(processes)])
     pipelineArgs.extend(["--register-dataset-types"])
 
+    # TODO: this config option resets the skymap dataset type to the templated
+    # name used by default prior to DM-27147.  After RFC-471 is adopted, the
+    # ap_verify test data should be migrated to use the new dataset type name
+    # (and collection naming convention) and this config override can be
+    # removed (other changes will probably be necessary as well, to reflect
+    # changes to the collection naming conventions).
+    pipelineArgs.extend(["--config", "imageDifference:connections.skyMap='{skyMapName}Coadd_skyMap'"])
```
kfindeisen, line 178: I think this will work, but since it doesn't depend on anything run-time, it might be cleaner to put it in `pipelines/ApVerify.py`.

kfindeisen (approved):

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -169,6 +169,14 @@ def runApPipeGen3(workspace, parsedCmdLine, processes=1):
     pipelineArgs.extend(["--processes", str(processes)])
     pipelineArgs.extend(["--register-dataset-types"])
 
+    # TODO: this config option resets the skymap dataset type to the templated
+    # name used by default prior to DM-27147.  After RFC-471 is adopted, the
+    # ap_verify test data should be migrated to use the new dataset type name
+    # (and collection naming convention) and this config override can be
+    # removed (other changes will probably be necessary as well, to reflect
+    # changes to the collection naming conventions).
```
kfindeisen, line 177: I suggest simplifying this to "TODO: remove after ap_verify datasets follow RFC-471 conventions".

python/lsst/ap/verify/pipeline_driver.py
TallJimbo: Good idea.  I'll test this first since it's in hand, then switch to that and make sure it still works.

TallJimbo: So, I _think_ that if I were to move the config override to the pipeline file, I'd have to duplicate the config overrides in the ap_pipe pipeline file, because a `task:` entry in an inherited pipeline replaces the inherited one instead of being merged with it (@natelust, can you confirm that?).  That makes me lean towards leaving this here, but I'm happy to follow your lead if you happen to look again tonight.

natelust: If the class pointed to by the label is the same in the base pipeline and the derived pipeline they are merged not replaced.

TallJimbo: Excellent, I'll go with to the pipeline solution, then.  Thanks!

PR #115, opened 2020-12-03
DM-26326: Remove obs dependency from ap_verify_testdata and ap_pipe_testdata

kfindeisen:
This PR compensates for the removal of the `obs_lsst` dependency from lsst/ap_verify_testdata#26 by adding a new dependency, and ensuring the `DataTestCase` base class checks separately for the testdata and the obs package being available.

### DIFF
--- a/python/lsst/ap/verify/testUtils.py
+++ b/python/lsst/ap/verify/testUtils.py
@@ -45,6 +45,12 @@ class DataTestCase(lsst.utils.tests.TestCase):
     testDataset = 'ap_verify_testdata'
     """The EUPS package name of the dataset to use for testing (`str`).
     """
+    obsPackage = 'obs_lsst'
+    """The obs package associated with ``testDataset`` (`str`).
+
+    Set to `None` if ``testDataset`` loads its own dependencies (not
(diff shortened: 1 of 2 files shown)


python/lsst/ap/verify/testUtils.py
```diff
@@ -45,6 +45,12 @@ class DataTestCase(lsst.utils.tests.TestCase):
     testDataset = 'ap_verify_testdata'
     """The EUPS package name of the dataset to use for testing (`str`).
     """
+    obsPackage = 'obs_lsst'
+    """The obs package associated with ``testDataset`` (`str`).
```
parejkoj, line 49: Is this approach really necessary? I guess this is future-proofing other subpackages/tests using different data?

parejkoj (commented):
One question, otherwise this looks good.

python/lsst/ap/verify/testUtils.py
kfindeisen: Pretty much, yes. It seemed silly (and hard to document) to support multiple datasets but require them to all be `obs_lsst`.

PR #118, opened 2020-12-18
DM-26593: Create an integrated pipeline for AP fake analysis

morriscb:


pipelines/ApVerifyWithFakes.yaml
```diff
@@ -0,0 +1,49 @@
+# Gen 3 pipeline for ap_verify
+# This concatenates various lsst.verify metrics to an AP pipeline
+
+description: Fully instrumented AP pipeline
+imports:
+    - location: $AP_PIPE_DIR/pipelines/ApPipe.yaml
+    - location: $AP_VERIFY_DIR/pipelines/MetricsRuntime.yaml
+    - location: $AP_VERIFY_DIR/pipelines/MetricsMiscWithFakes.yaml
```
kfindeisen, line 8: It might be more future-proof to import `ApVerify.yaml` here, though I see below that you end up overriding a lot of configs anyway.

pipelines/ApVerifyWithFakes.yaml
```diff
@@ -0,0 +1,49 @@
...
+    coaddFakes:
+        class: lsst.pipe.tasks.insertFakes.InsertFakesTask
+        config:
+            doSubSelectSources: True
+            sourceSelectionColName: "isTemplateSource"
+    visitFakes:
+        class: lsst.pipe.tasks.processCcdWithFakes.ProcessCcdWithFakesTask
+        config:
+            insertFakes.doSubSelectSources: True
+            insertFakes.sourceSelectionColName: "isVisitSource"
```
kfindeisen, line 26: Just to be sure: do you mean to have both this and regular CCD processing in the same pipeline?

pipelines/ApVerifyWithFakes.yaml
```diff
@@ -0,0 +1,49 @@
...
+    imageDifference:
+        class: lsst.pipe.tasks.imageDifference.ImageDifferenceTask
+        config:
+            # Always prefer decorrelation; may eventually become ImageDifferenceTask default
+            doDecorrelation: True
+            detection.thresholdValue: 5.0  # needed with doDecorrelation
+            # Don't have source catalogs for templates
+            doSelectSources: False
```
kfindeisen, line 34: You can leave these lines out; they're [redundant with the import](https://lsstc.slack.com/archives/C01FBUGM2CV/p1609187268233400).

pipelines/ApVerifyWithFakes.yaml
```diff
@@ -0,0 +1,49 @@
...
+            connections.fakesType: "fakes_"
+    diaPipe:
+        # TODO: how to prevent duplication with ApPipe definition?
+        class: lsst.ap.association.DiaPipelineTask
+        config:
+            apdb.isolation_level: READ_UNCOMMITTED
+            doPackageAlerts: True
```
kfindeisen, line 41: These lines are redundant if you import `ApVerify.yaml`, but not if you use `ApPipe.yaml`. (Putting the isolation level in the original pipeline was arguably a mistake, but that's out of scope for this ticket.)

pipelines/ApVerifyWithFakes.yaml
```diff
@@ -0,0 +1,49 @@
...
+            doWriteAssociatedSources: True
```
kfindeisen, line 42: I wasn't aware of this flag. Is it something we should be using in the main `ap_verify` pipeline?

pipelines/ApVerifyWithFakes.yaml
```diff
@@ -0,0 +1,49 @@
...
+            connections.fakesType: "fakes_"
+    matchFakes:
+        class: lsst.ap.pipe.matchApFakes.MatchApFakesTask
+contracts:
+    # Metric inputs must match pipeline outputs
+    - imageDifference.connections.coaddName == fracDiaSourcesToSciSources.connections.coaddName
+    - imageDifference.connections.fakesType == fracDiaSourcesToSciSources.connections.fakesType
```
kfindeisen, line 49: Redundant if you import `ApVerify.yaml`, but do you need any extra contracts for the `fakes*` tasks?

pipelines/MetricsMiscWithFakes.yaml
```diff
@@ -0,0 +1,44 @@
...
+        config:
+            connections.labelName: diaPipe
+    fracUpdatedDiaObjects:
+        class: lsst.ap.association.metrics.FractionUpdatedDiaObjectsMetricTask
+        config:
+            connections.labelName: diaPipe
+    totalUnassociatedDiaObjects:
+        class: lsst.ap.association.metrics.TotalUnassociatedDiaObjectsMetricTask
+    numSciSources:
+        class: lsst.ip.diffim.metrics.NumberSciSourcesMetricTask
```
kfindeisen, line 22: Why not replace these with an import? Or (given my advice to import `ApVerify.yaml` above) why not create a pipeline file that only has the fakes-specific metrics, then import it and `MetricsMisc.yaml` separately? As it stands, the files need to be kept in sync if somebody creates a new metric.

kfindeisen (approved):
Looks good, though there's a lot of code duplication that I think could be cleaned up.

pipelines/ApVerifyWithFakes.yaml
morriscb: Yes, `ProcessCcdWithFakesTask` needs the fully calibrated calexps to exist.

pipelines/ApVerifyWithFakes.yaml
morriscb: Not really. It's basically to write out the intermediate data product that ends up being written into the Apdb. Right now it's used pretty exclusively for the fakes task post processing steps. Might have another use later but not for now.

PR #121, opened 2021-03-23
DM-12549: ap_pipe must call AssociationTask in a reproducible order

kfindeisen:
This PR enforces the running of source association (or, more precisely, `DiaPipelineTask`) in Gen 3 in order of increasing visit ID. Association may still be run in parallel for different detectors within the same visit, and other tasks should be unaffected.

This PR makes no attempt to constrain the Gen 2 processing order, as the tooling needed to do so is completely different.

Depends on lsst/ctrl_mpexec#118.


andy-slac (approved):
Looks good.

PR #129, opened 2021-06-04
DM-29338: Incorporate Gen 3 crosstalk in HiTS runs

mrawls:
I successfully ran ap_verify on this ticket branch with the `ap_verify_ci_cosmos_pdr2` dataset.


isullivan (approved):
Looks fine, though I am curious why you are not using the `goodSeeing` templates here.

mrawls:
We could switch to using `goodSeeing` templates, but it would require more significant changes to modify the existing template-related data products in each ap_verify dataset. Perhaps something to consider in whatever form DM-28628 ever ends up taking.

PR #134, opened 2021-07-22
DM-31180: Force printing of pipeline logs.

kfindeisen:
This PR removes the `CliRunner` driver for Gen 3 `ap_verify` and replaces it with a direct `subprocess` call. (Looking back at the discussion that decided to use `CliRunner`, it appears `subprocess` was not even mentioned except as a joke...)

### DIFF
--- a/python/lsst/ap/verify/pipeline_driver.py
+++ b/python/lsst/ap/verify/pipeline_driver.py
@@ -152,6 +152,13 @@ def runApPipeGen3(workspace, parsedCmdLine, processes=1):
         Command-line arguments, including all arguments supported by `ApPipeParser`.
     processes : `int`
         The number of processes with which to call the AP pipeline
+
+    Returns
+    -------
+    code : `int`
+        An error code that is zero if the pipeline ran without problems, or
+        nonzero if there were errors. The exact meaning of nonzereo values
+        is an implementation detail.
     """
     log = lsst.log.Log.getLogger('ap.verify.pipeline_driver.runApPipeGen3')
 
--- a/python/lsst/ap/verify/pipeline_driver.py
+++ b/python/lsst/ap/verify/pipeline_driver.py
@@ -32,8 +32,7 @@ __all__ = ["ApPipeParser", "runApPipeGen2", "runApPipeGen3"]
 import argparse
 import os
 import re
-
-import click.testing
+import subprocess
 
 import lsst.log
 from lsst.utils import getPackageDir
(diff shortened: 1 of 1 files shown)


python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -175,16 +174,12 @@ def runApPipeGen3(workspace, parsedCmdLine, processes=1):
...
-        # CliRunner is an unsafe workaround for DM-26239
-        runner = click.testing.CliRunner()
+        # subprocess is an unsafe workaround for DM-26239
         # TODO: generalize this code in DM-26028
         # TODO: work off of workspace.workButler after DM-26239
-        results = runner.invoke(lsst.ctrl.mpexec.cli.pipetask.cli, pipelineArgs)
-        if results.exception:
-            raise RuntimeError("Pipeline failed.") from results.exception
-
+        result = subprocess.run(pipelineArgs, capture_output=False)
```
timj, line 180: in the previous implementation the pipetask run raised an exception on failure but this version returns an exit code (previously it raised but also returned an exit code). The docstring for this function does not say anything about what is returned. Can you update the docstring to be more explicit about intent here? If you still want it to raise I think you want `check=True` to be added here.

timj (approved):
Looks okay but there is a change of behavior that I'd like reflected in the docstring.

python/lsst/ap/verify/pipeline_driver.py
kfindeisen: Sorry, I'm confused. I know from testing that exceptions are propagated as-is, so the only behavior change there is the omission of one level of chaining. [Per the style guide](https://developer.lsst.io/python/numpydoc.html#py-docstring-raises), the exact exceptions raised by `runApPipeGen3` are not documented.

While `ctrl_mpexec` has been rewritten substantially since I first wrote this code, as far as I can tell there is still no documented guarantee that `pipetask run` or `CmdLineFwk.runPipeline` return any particular result. Therefore, formally I can't make any guarantee except that if there _is_ a meaningful return code, then it gets propagated.

timj: Okay. I'm also very confused. In the previous version of the code it had:

```
if results.exception:
    raise RuntimeError()
```
but in the current version of the code if pipetask fails the routine returns an exit code. Aren't those different behaviors? By default `subprocess.run()` does not raise an exception if the sub process returns bad exit status unless you also add `check=True`.

timj: The docstring doesn't tell me whether exit code is preferred over raising if there is a problem.

PR #135, opened 2021-07-29
DM-30349: Don't count fake sources in non-fakes metrics

kfindeisen:
This PR updates the `ApVerifyWithFakes` pipeline so that all metrics in `ip_diffim` and `pipe_tasks` can be expressed in terms of fakes-free datasets. This required some hacking to work around the default AP pipeline's assumptions about which tasks depended on which.

### DIFF
--- a/pipelines/ApVerifyWithFakes.yaml
+++ b/pipelines/ApVerifyWithFakes.yaml
@@ -4,6 +4,8 @@
 description: Fully instrumented AP pipeline with fakes
 imports:
     - location: $AP_VERIFY_DIR/pipelines/ApVerify.yaml
+      exclude:
+          - fracDiaSourcesToSciSources  # Selectively turn off this task's contracts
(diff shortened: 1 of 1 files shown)


pipelines/ApVerifyWithFakes.yaml
```diff
@@ -52,6 +76,7 @@ tasks:
             connections.diffIm: fakes_deepDiff_differenceExp
             connections.diaSourceTable: fakes_deepDiff_diaSrcTable
             # TODO: end DM-30210 workaround
+    # Can't have separate with- and without-fakes runs for diaPipe, because there's only one APDB
```
morriscb, line 79: I'm trying to think of a way we could have more than one Apdb for runs like this. Perhaps we could make a pipeline task with SkyMap as the only quantum, re-use a lot of our make_apdb code and then write a simple marker that is is the config data for the Apdb? Perhaps as a InputInit for the downstream tasks?

morriscb (approved):

PR #139, opened 2021-09-08
DM-26140: Centralize Gen 3 pipeline configuration info for ap_verify datasets

kfindeisen:
This PR adds instrument specializations of the `ApVerify` and `ApVerifyWithFakes` pipelines, and moves many of the previous pipeline overrides (in particular, the assumption of `deep` coadds) into individual datasets. It also modifies the behavior of the `ap_verify.py --pipeline` argument to more naturally support dataset-specific pipelines.


pipelines/ApVerifyWithFakes.yaml
```diff
@@ -126,6 +109,34 @@ tasks:
...
+            - createFakes
+            - coaddFakes
+        description: >
+            Creation and insertion of fake sources into data products.
+    apPipeWithFakes:
+        subset:
+            - isr
+            - characterizeImage
+            - calibrate
+            - visitFakes
```
mrawls, line 133: I do think this needs to be here, but it is odd to me that it's not in the `prepareFakes` subset. Maybe a comment explaining?

pipelines/ApVerifyWithFakes.yaml
```diff
@@ -126,6 +109,34 @@ tasks:
         config:
             connections.labelName: imageDifferenceNoFakes
             target: imageDifferenceNoFakes:measurement.run
+subsets:
+    apPipe:
+        subset:
+            - isr
+            - characterizeImage
+            - calibrate
+            - imageDifferenceNoFakes
```
mrawls, line 118: Where is `imageDifferenceNoFakes` coming from?

pipelines/DarkEnergyCamera/ApVerifyWithFakes.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+# This pipeline assumes $AP_PIPE_DIR/pipelines/DarkEnergyCamera/RunIsrForCrosstalkSources.yaml
+# has already been run.
+
+instrument: lsst.obs.decam.DarkEnergyCamera
+description: Fully instrumented AP pipeline with fakes, specialized for DECam
+imports:
+    - location: $AP_VERIFY_DIR/pipelines/ApVerifyWithFakes.yaml
+    # Do not use ap_pipe's ApPipe.yaml override, because ApVerifyWithFakes has
+    # *two* parallel AP pipelines.
```
mrawls, line 12: This comment is breaking my brain an little, what are the two parallel pipelines?

pipelines/HyperSuprimeCam/ApVerifyWithFakes.yaml
```diff
@@ -0,0 +1,9 @@
+# Gen 3 pipeline for ap_verify
+# This concatenates various lsst.verify metrics to an AP pipeline
+
+instrument: lsst.obs.subaru.HyperSuprimeCam
+description: Fully instrumented AP pipeline with fakes, specialized for HSC
+imports:
+    - location: $AP_VERIFY_DIR/pipelines/ApVerifyWithFakes.yaml
+    # Do not use ap_pipe's ApPipe.yaml override, because ApVerifyWithFakes has
+    # *two* parallel AP pipelines.
```
mrawls, line 9: Same request for clarity as with DECam

mrawls (commented):
As best I can tell these pipelines work how I'd expect, but I'm a bit lost in some of the details of the instrument-specific fakes pipelines. Please clarify, and make everything two-spaced indents while you're at it (sorry).

pipelines/ApVerifyWithFakes.yaml
kfindeisen: Just to make sure, is "this" just the `visitFakes` line? That probably is in the wrong place, based on me misunderstanding what that task is for.

pipelines/ApVerifyWithFakes.yaml
kfindeisen: It's defined higher up in the same file. (And yes, the awkward name is necessary to correctly deal with contracts inherited from `ApVerify.yaml`.)

pipelines/DarkEnergyCamera/ApVerifyWithFakes.yaml
kfindeisen: I meant that there's processing done with and without fakes, and aside from `visitFakes` depending on non-fakes `calibrate` they're independent of each other (I think -- this pipeline breaks my brain, too).

mrawls: Ah OK. Maybe just update this comment to say, e.g., "The ap_pipe package has a standard ApPipe.yaml and a parallel one that includes fake/synthetic sources. We are intentionally not using the former here."

pipelines/ApVerifyWithFakes.yaml
natelust: I dont really know what each of these tasks is used for, but as a note, it is possible to include a label in more than one subset, they do not have to be mutually exclusive. In DRP there is processCcd, but also Single frame that includes even more. These are just helpful organizers for people to help run or refer to collections of tasks.

kfindeisen: I actually did mean to create a partition, because that's the organization that makes the most sense to me.

And on third(?) thought, I think `visitFakes` belongs in `apPipeWithFakes` (because it runs  `characterizeImage` and `calibrate` on the data) but not in `prepareFakes` (because it requires partially processed raws whereas the other tasks can be done without them). Not ready to push yet, but here's what I have so far:

```
  apPipe:
    subset:
      - isr
      - characterizeImage
      - calibrate
      - imageDifferenceNoFakes
    description: >
      The AP pipeline without fakes. Only includes processing through
      image differencing.
  prepareFakes:
    subset:
      - createFakes
      - coaddFakes
    description: >
      Creation of fake sources.
  apPipeWithFakes:
    subset:
      - visitFakes    # characterizeImage and calibrate with fakes
      - imageDifference
      - transformDiaSrcCat
      - diaPipe
      - matchFakes
    description: >
      The AP pipeline with fakes. Requires apPipe and prepareFakes subsets.
```

PR #143, opened 2021-11-08
DM-31801: Use lsst prefix for loggers

timj:
I think in order to switch to python logging we'll have to add a logging.basicConfig and lsst.log forwarding in here somewhere.

### DIFF
--- a/python/lsst/ap/verify/ap_verify.py
+++ b/python/lsst/ap/verify/ap_verify.py
@@ -41,6 +41,8 @@ from .metrics import MetricsParser, computeMetrics
 from .pipeline_driver import ApPipeParser, runApPipeGen2, runApPipeGen3
 from .workspace import WorkspaceGen2, WorkspaceGen3
 
+_LOG = lsst.log.Log.getLogger(__name__)
+
 
 class _InputOutputParser(argparse.ArgumentParser):
     """An argument parser for program-wide input and output.
@@ -188,7 +190,7 @@ def runApVerify(cmdLine=None):
         or 127 if the task runner framework failed.
     """
     lsst.log.configure()
-    log = lsst.log.Log.getLogger('ap.verify.ap_verify.main')
+    log = _LOG.getChild('main')
     # TODO: what is LSST's policy on exceptions escaping into main()?
     args = _ApVerifyParser().parse_args(args=cmdLine)
     log.debug('Command-line arguments: %s', args)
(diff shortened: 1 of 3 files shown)


python/lsst/ap/verify/ingestion.py
```diff
@@ -48,6 +48,8 @@
 from lsst.pipe.tasks.ingestCalibs import IngestCalibsTask
 from lsst.pipe.tasks.ingestCuratedCalibs import IngestCuratedCalibsTask
 
+_LOG = lsst.log.Log.getLogger(__name__)
```
kfindeisen, line 51: Do we need this boilerplate in every module? I'm a bit suspicious of global variables...

python/lsst/ap/verify/ap_verify.py
```diff
@@ -188,7 +190,7 @@ def runApVerify(cmdLine=None):
         or 127 if the task runner framework failed.
     """
     lsst.log.configure()
```
kfindeisen, line 192: I believe this call and the one in `runIngestion` are what should be replaced with any `logging.basicConfig` and `lsst.log` forwarding that is needed. These two functions should be the only (non-test) entry points to the code.

kfindeisen (approved):
I'm a bit unclear about the scope of this work, but have no serious objections to the changes.

python/lsst/ap/verify/ingestion.py
timj: Where else would you put it? We use this form all the time in middleware.

timj:
@kfindeisen I switched it to python logging. I'm not sure what default log configuration you'd like but I use INFO and log messages to stdout.

timj:
The scope of this work is to change the logger names but there is also a general requirement to switch all python code over to python logging where practical. Since I was already in the files I thought I may as well do that on this ticket.

python/lsst/ap/verify/ap_verify.py
```diff
@@ -41,7 +43,17 @@
...
+
+
+def _configure_logger():
+    """Configure Python logging.
+
+    Does basic Python logging configuration and
+    forwards LSST logger to Python logging.
+    """
+    logging.basicConfig(level=logging.INFO, stream=sys.stdout)
+    lsst.log.configure_pylog_MDC("DEBUG", MDC_class=None)
```
kfindeisen, line 56: No objection to `INFO`, but what is this last line?

kfindeisen: Never mind, it's obvious.

kfindeisen (approved):

python/lsst/ap/verify/ap_verify.py
timj: That forwards all DEBUG and above messages to python logging -- python will drop the actual DEBUG messages. MDC is disabled.

PR #148, opened 2022-01-26
DM-33150: Remove --gen flags from CI bridge script.

kfindeisen:
This PR stops `run_ci_dataset.sh` from using obsolete `ap_verify.py` flags. This change should have been made as part of #147, where these flags were removed.

The `-g` argument to `run_ci_dataset.sh` is kept, in case we find a need to support something similar in the future.

### DIFF
--- a/bin/run_ci_dataset.sh
+++ b/bin/run_ci_dataset.sh
@@ -39,8 +39,8 @@ usage() {
     print_error
     print_error "Specific options:"
     print_error "   -d          Dataset name"
-    print_error "   -g          Middleware generation number (int)"
-    print_error "   -p          Pipeline to run (Gen 3 only)"
+    print_error "   -g          Middleware generation number (int) [currently unused]"
+    print_error "   -p          Pipeline to run"
     print_error "   -h          show this message"
     exit 1
 }
@@ -58,9 +58,6 @@ if [[ -z "${DATASET}" ]]; then
     print_error "$0: mandatory argument -- d"
     usage
     exit 1
-fi
-if [[ -n "${GEN}" ]]; then
-    GEN="--gen${GEN}"
 fi
 if [[ -n "${PIPE}" ]]; then
     PIPE="--pipeline ${PIPE}"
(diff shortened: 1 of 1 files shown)


bin/run_ci_dataset.sh
```diff
@@ -90,11 +87,9 @@ fi
...
 NUMPROC=${NUMPROC:-$((sys_proc < max_proc ? sys_proc : max_proc))}
 
-echo "Running ap_verify on ${DATASET} ${GEN}..."
+echo "Running ap_verify on ${DATASET}..."
 ap_verify.py --dataset "${DATASET}" \
-    ${GEN} \
     ${PIPE} \
     --output "${WORKSPACE}" \
     --processes "${NUMPROC}" \
-    --metrics-file "${WORKSPACE}/ap_verify.{dataId}.verify.json" \
```
parejkoj, line 99: Why did this line also go?

kfindeisen: This argument was Gen 2-only, and was removed on c145de5eb67490e1df7a5708450e7de6a9b9777f.

bin/run_ci_dataset.sh
```diff
@@ -39,8 +39,8 @@ usage() {
     print_error
     print_error "Specific options:"
     print_error "   -d          Dataset name"
-    print_error "   -g          Middleware generation number (int)"
-    print_error "   -p          Pipeline to run (Gen 3 only)"
+    print_error "   -g          Middleware generation number (int) [currently unused]"
```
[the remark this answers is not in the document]
kfindeisen: Nothing's been missed -- the Jenkins script [still calls this argument](https://github.com/lsst-dm/jenkins-dm-jobs/blob/1aa272fe6cda2e557831dc1e0245cdb424ee8e06/pipelines/scipipe/ap_verify.groovy#L429).

The main reasons I don't want to remove it completely are

a) this would require unraveling a great deal of Jenkins code, all the way down to the config format and everything that depends on *that*
b) a lot of how to support multiple Butler generations on the Jenkins side took a long time to work out, and I'd rather keep it around (despite the triviality of `gen === 3`) than have to reinvent it when we have a similar problem.

PR #150, opened 2022-01-28
DM-33440: need to fix ap_verify for changes in DM-32988

kherner:


pipelines/ApVerifyWithFakes.yaml
```diff
@@ -71,10 +71,12 @@ tasks:
       connections.warpedExposure: parameters.diffWarp
       connections.matchedExposure: parameters.diffMatch
       connections.diaSources: parameters.diaSrcCat
+      connections.inputTemplate: parameters.templateExp
       # TODO: end DM-30210 workaround
   imageDifference:
-    class: lsst.pipe.tasks.imageDifference.ImageDifferenceTask
+    class: lsst.pipe.tasks.imageDifference.ImageDifferenceFromTemplateTask
     config:
+      doWriteWarpedExp: True       # Required for packaging alerts in diaPipe
```
kfindeisen, line 79: This line should not be necessary, but I'm fine with leaving it for symmetry.

kfindeisen (approved):
Thanks for the quick fix!

pipelines/ApVerifyWithFakes.yaml
kherner: Yes, I thought it could lead to possible confusion for a new person somewhere down the line if it was in one and not the other.

PR #156, opened 2022-03-30
DM-34237: Fix subsets in ApVerifyWithFakes pipeline

kfindeisen:
This PR fixes a minor bug introduced in #155.


isullivan (approved):
Thank you for catching and fixing my mistake!

PR #157, opened 2022-04-20
DM-27013: Change CalibrateTask refcat defaults to Gaia DR2 for astrometry and PS1 for photometry

parejkoj:


kfindeisen (approved):
Changes look good. This is unfortunately a weakness of these mocks...

PR #159, opened 2022-05-04
DM-32694: Split AP pipeline into ApPipeWithFakes

kfindeisen:
This PR refactors the `ApVerifyWithFakes` pipeline to depend on `ApPipeWithFakes` (itself reworked on lsst/ap_pipe#106) instead of `ApVerify`. This change improves consistency with the "base" AP pipeline and avoids some of the contract conflicts that made the previous `ApVerifyWithFakes` more complicated and more brittle than it needed to be.

### DIFF
--- a/pipelines/ApVerify.yaml
+++ b/pipelines/ApVerify.yaml
@@ -13,5 +13,6 @@ tasks:
       doPackageAlerts: True
 contracts:
   # Metric inputs must match pipeline outputs
-  - imageDifference.connections.coaddName == fracDiaSourcesToSciSources.connections.coaddName
-  - imageDifference.connections.fakesType == fracDiaSourcesToSciSources.connections.fakesType
+  # Use of ConnectionsClass for templated fields is a workaround for DM-30210
+  - imageDifference.connections.ConnectionsClass(config=imageDifference).diaSources.name ==
+      fracDiaSourcesToSciSources.connections.ConnectionsClass(config=fracDiaSourcesToSciSources).diaSources.name
--- a/pipelines/ApVerifyWithFakes.yaml
+++ b/pipelines/ApVerifyWithFakes.yaml
@@ -3,9 +3,9 @@
 
 description: Fully instrumented AP pipeline with fakes
 imports:
-  - location: $AP_VERIFY_DIR/pipelines/ApVerify.yaml
-    exclude:
-      - fracDiaSourcesToSciSources  # Selectively turn off this task's contracts
+  - location: $AP_PIPE_DIR/pipelines/ApPipeWithFakes.yaml
+  - location: $AP_VERIFY_DIR/pipelines/MetricsRuntime.yaml
+  - location: $AP_VERIFY_DIR/pipelines/MetricsMisc.yaml
   - location: $AP_VERIFY_DIR/pipelines/MetricsForFakes.yaml
 parameters:
   fakesType: fakes_
@@ -24,12 +24,16 @@ tasks:
       select_col: "isTemplateSource"
       connections.coaddName: parameters.coaddName
       connections.fakesType: parameters.fakesType
-  singleFrameWithFakes:
+  processVisitFakes:
     class: lsst.pipe.tasks.processCcdWithFakes.ProcessCcdWithVariableFakesTask
     config:
       connections.fakesType: parameters.fakesType
       insertFakes.doSubSelectSources: True
       insertFakes.select_col: "isVisitSource"
+  retrieveTemplate:
+    class: lsst.ip.diffim.getTemplate.GetTemplateTask
+    config:
+        connections.coaddName: parameters.coaddName
   retrieveTemplateWithFakes:
     class: lsst.ip.diffim.getTemplate.GetTemplateTask
     config:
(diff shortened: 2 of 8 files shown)


pipelines/ApVerify.yaml
```diff
@@ -13,5 +13,6 @@ tasks:
       doPackageAlerts: True
 contracts:
   # Metric inputs must match pipeline outputs
-  - imageDifference.connections.coaddName == fracDiaSourcesToSciSources.connections.coaddName
-  - imageDifference.connections.fakesType == fracDiaSourcesToSciSources.connections.fakesType
+  # Use of ConnectionsClass for templated fields is a workaround for DM-30210
```
mrawls, line 16: I like this revised workaround, great thinking

pipelines/ApVerifyWithFakes.yaml
```diff
@@ -3,211 +3,51 @@
...
-      connections.fakeCat: parameters.fakesInput
-      connections.imageWithFakes: parameters.fakesTemplate
-      # TODO: end DM-30210 workaround
-  singleFrameWithFakes:
-    class: lsst.pipe.tasks.processCcdWithFakes.ProcessCcdWithVariableFakesTask
-    config:
-      insertFakes.doSubSelectSources: True
-      insertFakes.select_col: "isVisitSource"
-  retrieveTemplateWithFakes:
+  # Parallel to retrieveTemplateWithFakes, allows non-fakes image differencing.
```
mrawls, line 11: Here and below, please clarify in the comment that you're referring to how ApPipe labels the task (it took me a minute to figure out what it was parallel to)

pipelines/ApVerifyWithFakes.yaml
```diff
@@ -3,211 +3,51 @@
...
-      connections.fakesType: parameters.fakesType
-      # TODO: redundant connection definitions workaround for DM-30210
-      connections.coaddName: parameters.coaddName
-      connections.diaSourceSchema: parameters.fakesDiaSrcSchema
-      connections.diaSourceCat: parameters.fakesDiaSrcCat
-      connections.diffIm: parameters.fakesDiff
-      connections.diaSourceTable: parameters.fakesDiaSrcParquet
-      # TODO: end DM-30210 workaround
-  # Can't have separate with- and without-fakes runs for diaPipe, because there's only one APDB
+
```
mrawls, line 24: Remove blank line?

pipelines/HyperSuprimeCam/ApVerifyWithFakes.yaml
```diff
@@ -5,6 +5,6 @@ instrument: lsst.obs.subaru.HyperSuprimeCam
 description: Fully instrumented AP pipeline with fakes, specialized for HSC
 imports:
```
mrawls, line 6: Why does the DECam version of this pipeline have a whole separate `processVisitFakes` section, but this HSC one does not?

mrawls (approved):
Minor comments only, basically looks good, thank you!

pipelines/HyperSuprimeCam/ApVerifyWithFakes.yaml
kfindeisen: Good question... those configs were added on https://github.com/lsst/ap_verify/commit/801f904a44fde924046e860e977dcd5debc8dad9 without explanation. I think they might have been cargo-culted from the `calibrate` config in https://github.com/lsst/ap_pipe/commit/1c8f5877dee5610a62ffc02ffa25072d70dab025. There is no analogous set of configs for `calibrate` in the HSC `ProcessCcd.yaml`.

mrawls: Oh, I see, OK.

pipelines/DarkEnergyCamera/ApVerifyWithFakes.yaml
```diff
@@ -24,7 +24,7 @@ tasks:
     config:
       photoCal.match.referenceSelection.magLimit.fluxField: i_flux
       photoCal.match.referenceSelection.magLimit.maximum: 22.0
-  singleFrameWithFakes:
+  processVisitFakes:
```
mrawls, line 27: Would it be possible to delete this section so it pulls that calibrate config in from elsewhere? It's the default in `$AP_PIPE_DIR/pipelines/DarkEnergyCamera/ProcessCcd.yaml`, so maybe it is already properly imported as the default here, too?

kfindeisen: I could (should?) move it to an `$AP_PIPE_DIR/pipelines/DarkEnergyCamera/ApPipeWithFakes.yaml`, but it would still be asymmetric with HSC. But I don't understand what you mean by it being the default; `ProcessCcd.yaml` doesn't know about `processVisitFakes`.

mrawls: OK, scratch that, I forgot that `processVisitFakes` is code for "insert fakes into calexps and redo characterize and calibrate for reasons." Therefore, any camera-specific ProcessCcd configs *do* need to explicitly be here.

pipelines/HyperSuprimeCam/ApVerifyWithFakes.yaml
mrawls: Nevermind, per above DECam comment, there are no special characterize or calibrate configs in HSC's ProcessCcd pipeline, so we don't need to have any config overrides here either.

pipelines/DarkEnergyCamera/ApVerifyWithFakes.yaml
kfindeisen: Also scratch what I said, we specifically _don't_ want to import `ApPipeWithFakes` because of conflicts with multiple overrides to the same task.

However, @mrawls recommended on Slack that the hardcoded `isr` and `calibrate` overrides here be replaced with an import of instrument-specific `ProcessCcd`; this won't conflict with base `ApVerifyWithFakes` because the latter doesn't touch those tasks at all.

PR #161, opened 2022-06-02
DM-35071: Add support for datasets using old Gen 3 registry versions.

kfindeisen:
This PR allows `ap_verify` to read datasets whose `preloaded` repo has a registry version other than the latest. This solution involves using an unsafe keyword for `RawIngestTask`, but avoids dataset churn.

This fix is not unit tested, since `ap_verify_testdata` does not have solar system data and therefore does not export exposure records.


timj (approved):
Looks okay to me.

PR #162, opened 2022-06-24
DM-34801: Replace Diff_warpedExp with Diff_templateExp in diaPipe

parejkoj:


tests/MockApPipe.yaml
```diff
@@ -12,7 +12,7 @@ parameters:
   diaSrcParquet: goodSeeingDiff_diaSrcTable
   diff: goodSeeingDiff_differenceExp
   diffScore: goodSeeingDiff_scoreExp
-  diffWarp: goodSeeingDiff_warpedExp
+  diffWarp: goodSeeingDiff_templateExp
   diffMatch: goodSeeingDiff_matchedExp
   assocSrc: goodSeeingDiff_assocDiaSrc
   # TODO: end DM-30210 workaround
```
kfindeisen: I think ce5919a can be dropped, since d7554bb removes these blocks entirely.

tests/MockApPipe.yaml
```diff
@@ -5,17 +5,6 @@ parameters:
...
-  template: goodSeeingCoadd
-  diaSrcCat: goodSeeingDiff_diaSrc
-  diaSrcSchema: goodSeeingDiff_diaSrc_schema
-  diaSrcParquet: goodSeeingDiff_diaSrcTable
-  diff: goodSeeingDiff_differenceExp
-  diffScore: goodSeeingDiff_scoreExp
-  diffWarp: goodSeeingDiff_templateExp
-  diffMatch: goodSeeingDiff_matchedExp
-  assocSrc: goodSeeingDiff_assocDiaSrc
-  # TODO: end DM-30210 workaround
```
kfindeisen, line 18: Many thanks, I'm embarrassed to have missed this!

tests/MockApPipe.yaml
```diff
@@ -31,6 +31,7 @@ tasks:
       doSkySources: True
       coaddName: parameters.coaddName              # Can be removed once ImageDifference no longer supports Gen 2
       connections.coaddName: parameters.coaddName
+      connections.warpedExposure: goodSeeingDiff_templateExp
```
kfindeisen, line 34: I think this line can be removed after the correction below.

tests/MockApPipe.yaml
```diff
@@ -50,16 +51,14 @@ contracts:
   - imageDifference.doSkySources == transformDiaSrcCat.doRemoveSkySources
   # Inputs and outputs must match
   # Use of ConnectionsClass for templated fields is a workaround for DM-30210
-  - retrieveTemplate.connections.ConnectionsClass(config=retrieveTemplate).outputExposure.name ==
-      imageDifference.connections.ConnectionsClass(config=imageDifference).inputTemplate.name
```
kfindeisen: Why is this contract invalid? [The real pipeline](https://github.com/lsst/ap_pipe/blob/main/pipelines/ApPipe.yaml#L60-L61) uses it.

tests/MockApPipe.yaml
```diff
@@ -50,16 +51,14 @@ contracts:
...
   - imageDifference.connections.ConnectionsClass(config=imageDifference).subtractedExposure.name ==
       transformDiaSrcCat.connections.ConnectionsClass(config=transformDiaSrcCat).diffIm.name
   - imageDifference.connections.ConnectionsClass(config=imageDifference).diaSources.name ==
       transformDiaSrcCat.connections.ConnectionsClass(config=transformDiaSrcCat).diaSourceCat.name
   - imageDifference.connections.ConnectionsClass(config=imageDifference).subtractedExposure.name ==
         diaPipe.connections.ConnectionsClass(config=diaPipe).diffIm.name
   - imageDifference.connections.ConnectionsClass(config=imageDifference).warpedExposure.name ==
-        diaPipe.connections.ConnectionsClass(config=diaPipe).warpedExposure.name
+        diaPipe.connections.ConnectionsClass(config=diaPipe).template.name
```
kfindeisen, line 61: Shouldn't you rename both terms, as you did in `ap_pipe`?

kfindeisen (changes requested):
I think the changes to `MockApPipe.yaml` got a bit muddled; they're not analogous to the changes made on lsst/ap_pipe#115.

tests/MockApPipe.yaml
parejkoj: There is no `retrieveTemplate` in the mock pipeline.

kfindeisen (approved):

tests/MockApPipe.yaml
kfindeisen: So, to be fixed when we swap out `imageDifference`?

tests/MockApPipe.yaml
parejkoj: No, I need to keep some of the ones in the mock pipeline, because it's based on `ImageDifferenceTask`, not `ImageDifferenceFromTemplateTask`. We'll want to rework these to use the new SubtractTask anyway.

PR #163, opened 2022-06-28
DM-34445 Update ap_verify tutorial

AndyTza:


ebellm (approved):
Looks great!

doc/lsst.ap.verify/running.rst
```diff
@@ -17,18 +17,27 @@ For more details, see the :doc:`command-line-reference` or run :option:`ap_verif
...
-.. _Cosmos PDR2: https://github.com/lsst/ap_verify_ci_cosmos_pdr2/
+.. _Cosmos PDR2: https://github.com/lsst/ap_verify_ci_cosmos_pdr2/ 
+
+First download and setup the dataset. 
+
+.. prompt:: bash
+
+   git clone https://github.com/lsst/ap_verify_ci_cosmos_pdr2/
+   setup -r ap_verify_ci_cosmos_pdr2
+
```
kfindeisen, line 30: This is incorrect. `ap_verify` datasets should only be downloaded once, and not every time `ap_verify` is run; they can be very large!

If you believe it is necessary to mention installation, I would recommend linking to the installation docs from this page's introduction. This practice is unfortunately not in the developer guide, but see [DMTN-030](https://dmtn-030.lsst.io/#topic-based-documentation).

PR #164, opened 2022-07-13
DM-35564: ap_verify docs say datasets should be installed at run-time

kfindeisen:
This PR cleans up the docs added in #163 to clearly distinguish between dataset installation, dataset setup, and running `ap_verify`.


mrawls (approved):
Good clarifications, thank you

PR #165, opened 2022-07-26
DM-35639: Switch to use the new image differencing

isullivan:


pipelines/MetricsRuntime.yaml
```diff
@@ -24,54 +24,38 @@ tasks:
...
-      connections.labelName: imageDifference
-      metadataDimensions: [instrument, visit, detector, skymap]
-      target: imageDifference:register.run
-  timing_imageDifference_subtract:
+      connections.labelName: subtractImages
+      metadataDimensions: [instrument, visit, detector]
+      target: subtractImages.run
     class: lsst.verify.tasks.commonMetrics.TimingMetricTask
+  timing_detectAndMeasure:
     config:
```
kfindeisen, line 37: There's no `class` key here, which breaks the pipeline. Please fix and confirm that the pipeline works.

pipelines/MetricsRuntime.yaml
```diff
@@ -24,54 +24,38 @@ tasks:
...
       connections.package: ip_diffim
-      connections.metric: ImagePsfMatchTime
-      connections.labelName: imageDifference
-      metadataDimensions: [instrument, visit, detector, skymap]
-      target: imageDifference:subtract.subtractExposures
-  timing_imageDifference_detection:
+      connections.metric: DetectAndMeasureTime
+      connections.labelName: detectAndMeasure
+      metadataDimensions: [instrument, visit, detector]
+      target: detectAndMeasure.run
```
kfindeisen, line 42: It looks like `ip_diffim.DetectAndMeasureTime` is supposed to be a new metric, but I don't see a PR in `verify_metrics`. Even if the pipeline runs, I don't think SQuaSH upload will work without it.

It might be a good chance to get rid of `AstrometryTime` and `RegisterImageTime` as well, since I don't think they were ever used.

kfindeisen (changes requested):
Unfortunately, I can't get `ap_verify` to run with these changes, and it looks like you tried to use new metrics without creating them first.

(I'll still do a proper review, but I thought it would be better to let you know about these two issues up front.)

pipelines/ApVerifyWithFakes.yaml
```diff
@@ -14,13 +14,16 @@ tasks:
     config:
         connections.coaddName: parameters.coaddName
   # Parallel to ApPipe's imageDifferenceWithFakes, allows clean differencing metrics.
-  imageDifference:
-    class: lsst.pipe.tasks.imageDifference.ImageDifferenceFromTemplateTask
+  subtractImages:
+    class: lsst.ip.diffim.subtractImages.AlardLuptonSubtractTask
+    config:
+      connections.coaddName: parameters.coaddName
+      doApplyFinalizedPsf: False
```
kfindeisen, line 21: Is this line in case `ApVerifyWithFakes` some day uses a different subtracter from `ApPipeWithFakes`, preventing config inheritance? If so, a comment would be helpful.

pipelines/MetricsRuntime.yaml
```diff
@@ -24,54 +24,38 @@ tasks:
       connections.metric: CalibrateTime
       connections.labelName: calibrate
       target: calibrate.run
-  timing_imageDifference:
+  timing_subtractImages:
     class: lsst.verify.tasks.commonMetrics.TimingMetricTask
     config:
-      connections.package: pipe_tasks
+      connections.package: ip_diffim
       connections.metric: ImageDifferenceTime
```
kfindeisen, line 31: Renaming `pipe_tasks.ImageDifferenceTime` to `ip_diffim.ImageDifferenceTime` needs to be done in `verify_metrics`.

If you're renaming the metric anyway, would it be better to call it `SubtractImagesTime`? Our SQuaSH graphs will need to be modified either way (and that may be tricky, since the namespace is fundamental to how metrics are organized in the online database 😬).

pipelines/MetricsRuntime.yaml
```diff
@@ -24,54 +24,38 @@ tasks:
...
-    class: lsst.verify.tasks.commonMetrics.TimingMetricTask
-    config:
-      connections.package: pipe_tasks
-      connections.metric: RegisterImageTime
-      connections.labelName: imageDifference
-      metadataDimensions: [instrument, visit, detector, skymap]
-      target: imageDifference:register.run
-  timing_imageDifference_subtract:
+      connections.labelName: subtractImages
+      metadataDimensions: [instrument, visit, detector]
```
kfindeisen, line 33: Visit-detector are the default dimensions for `MetadataMetricTask`, so this override is unnecessary, and it's better not to use that hack if we don't have to. Same with the other metrics you've added.

kfindeisen (changes requested):
The core pipelines look good, but metrics have some issues.

pipelines/ApVerifyWithFakes.yaml
isullivan: No, this change is unneeded. I'll remove it.

pipelines/MetricsRuntime.yaml
isullivan: I'll make the `verify_metrics` PR. What else is needed for these metrics to be picked up by SQuaSH?

kfindeisen: That should be it.

pipelines/MetricsRuntime.yaml
```diff
@@ -24,54 +24,34 @@ tasks:
...
-    class: lsst.verify.tasks.commonMetrics.TimingMetricTask
-    config:
-      connections.package: pipe_tasks
-      connections.metric: RegisterImageTime
-      connections.labelName: imageDifference
-      metadataDimensions: [instrument, visit, detector, skymap]
-      target: imageDifference:register.run
-  timing_imageDifference_subtract:
+      connections.package: ip_diffim
+      connections.metric: SubtractImagesTime
```
kfindeisen, line 31: This should be `SubtractImagesALTime`; did you forget to push?

kfindeisen (changes requested):

pipelines/MetricsRuntime.yaml
isullivan: Oops! Pushed now.

kfindeisen (approved):

PR #169, opened 2022-09-15
DM-36198: Add parquet transform tasks to ap_verify

kfindeisen:
This PR adds extra Parquet transformation tasks to both `ap_verify` pipelines. These tasks are not needed (or used) in CI, but make it easier to make plots of `ap_verify` datasets in `analysis_tools`.

### DIFF
--- a/pipelines/ConversionsForFakes.yaml
+++ b/pipelines/ConversionsForFakes.yaml
@@ -1,0 +1,54 @@
+# Add-on pipeline designed to be added to ApVerifyWithFakes or ApPipeWithFakes
+# for QA purposes.
+#
+# This pipeline depends on an external pipeline for the coaddName and fakesType
+# pipeline parameters.
+
+description: Type conversion tasks customized for AP pipeline
+imports:
+  - location: $AP_VERIFY_DIR/pipelines/Conversions.yaml
+    exclude:
+      # Fakes pipeline doesn't produce non-fakes DiaSourceTable.
+      - consolidateDiaSourceTable
+tasks:
+  # Conversion of fakes_src [afw.table] to fakes_source [Parquet]
+  writeSourceTableWithFakes:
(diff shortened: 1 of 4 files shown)


pipelines/ConversionsForFakes.yaml
```diff
@@ -0,0 +1,54 @@
...
+  - location: $AP_VERIFY_DIR/pipelines/Conversions.yaml
+    exclude:
+      # Fakes pipeline doesn't produce non-fakes DiaSourceTable.
+      - consolidateDiaSourceTable
+tasks:
+  # Conversion of fakes_src [afw.table] to fakes_source [Parquet]
+  writeSourceTableWithFakes:
+    class: lsst.pipe.tasks.postprocess.WriteSourceTableTask
+    config:
+      connections.catalogType: parameters.fakesType
```
parejkoj, line 18: I'm sad that there's not a way to override all of the e.g. `catalogType` fields at once, but so be it.

parejkoj (approved):
Thank you for all of the explanatory comments in the yaml files: it's helpful for understanding what's going on.

You say in the PR description that these are not used or needed in CI, but won't we eventually be using these to generate plots and metrics in the nightly CI runs? I don't think that affects the PR contents, though.

pipelines/ConversionsForFakes.yaml
kfindeisen: We actually don't want to do that, because we have non-fakes versions of these tasks in the assembled `ApPipeWithFakes`. They need to have different values for their templates.

kfindeisen:
> Won't we eventually be using these to generate plots and metrics in the nightly CI runs

Last I asked the bosses, it was not clear whether and in what way we'd be using `analysis_tools`. They said part of the point of the sprint was to figure that out.

PR #174, opened 2022-11-24
DM-36743: Add the real-bogus PipelineTask to ap_verify for the DC2 CI dataset

NimSed:
plugged rbClassify into dc2 ap_verify

### DIFF
--- a/pipelines/LSSTCam-imSim/ApVerify.yaml
+++ b/pipelines/LSSTCam-imSim/ApVerify.yaml
@@ -18,7 +18,7 @@ tasks:
   rbClassify:
     class: lsst.meas.transiNet.TransiNetTask
     config:
-      modelPackageName: 'neighbor:///rbResnet50-DC2'
+      modelPackageName: 'rbResnet50-DC2'
       connections.coaddName: parameters.coaddName
 contracts:
   # Must re-declare contracts that cross apPipe and metrics boundary, as
--- a/pipelines/LSSTCam-imSim/ApVerify.yaml
+++ b/pipelines/LSSTCam-imSim/ApVerify.yaml
@@ -18,8 +18,7 @@ tasks:
   rbClassify:
     class: lsst.meas.transiNet.TransiNetTask
     config:
-      modelFile: testModel
-      weightsFile: '/home/nima/meas_transiNet/tests/data/checkpoint.pth.zip'
+      modelPackageName: 'neighbor:///rbResnet50-DC2'
       connections.coaddName: parameters.coaddName
 contracts:
   # Must re-declare contracts that cross apPipe and metrics boundary, as
--- a/pipelines/LSSTCam-imSim/ApVerify.yaml
+++ b/pipelines/LSSTCam-imSim/ApVerify.yaml
@@ -20,7 +20,7 @@ tasks:
     config:
       modelFile: testModel
       weightsFile: '/home/nima/meas_transiNet/tests/data/checkpoint.pth.zip'
-      connections.coaddName: 'goodSeeing'
+      connections.coaddName: parameters.coaddName
 contracts:
   # Must re-declare contracts that cross apPipe and metrics boundary, as
   # these were removed on import.
--- a/pipelines/LSSTCam-imSim/ApVerify.yaml
+++ b/pipelines/LSSTCam-imSim/ApVerify.yaml
@@ -15,6 +15,12 @@ tasks:
     class: lsst.ap.association.DiaPipelineTask
     config:
       doPackageAlerts: True
+  rbClassify:
+    class: lsst.meas.transiNet.TransiNetTask
+    config:
+      modelFile: testModel
+      weightsFile: '/home/nima/meas_transiNet/tests/data/checkpoint.pth.zip'
+      connections.coaddName: 'goodSeeing'
 contracts:
   # Must re-declare contracts that cross apPipe and metrics boundary, as
   # these were removed on import.
--- a/pipelines/LSSTCam-imSim/ApVerify.yaml
+++ b/pipelines/LSSTCam-imSim/ApVerify.yaml
@@ -18,8 +18,6 @@ tasks:
   transformDiaSrcCat:
     class: lsst.ap.association.TransformDiaSourceCatalogTask
     config:
-      doRemoveSkySources: True
-      connections.coaddName: parameters.coaddName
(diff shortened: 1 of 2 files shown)


pipelines/LSSTCam-imSim/ApVerify.yaml
```diff
@@ -15,6 +15,12 @@ tasks:
     class: lsst.ap.association.DiaPipelineTask
     config:
       doPackageAlerts: True
+  rbClassify:
+    class: lsst.meas.transiNet.TransiNetTask
+    config:
+      modelFile: testModel
+      weightsFile: '/home/nima/meas_transiNet/tests/data/checkpoint.pth.zip'
```
kfindeisen, line 22: I don't think this will work -- `ap_verify` datasets must be completely self-contained (in particular, this path won't be visible from Jenkins). Is this "weights file" something that can be added to the Butler repository, like solar system ephemerides were for the HiTS and COSMOS datasets?

pipelines/LSSTCam-imSim/ApVerify.yaml
```diff
@@ -15,6 +15,12 @@ tasks:
...
+      connections.coaddName: 'goodSeeing'
```
kfindeisen, line 23: Hard-coding the coadd name causes a lot of problems when pipelines must be customized for specific datasets:
```suggestion
      connections.coaddName: parameters.coaddName
```

pipelines/LSSTCam-imSim/ApVerify.yaml
NimSed: This is the discussion I was pointing at during today's meeting. We are aware of the need to ingest these big chunks of data into the repository. But the question is how. There is https://jira.lsstcorp.org/browse/DM-36742 where I think your comments are needed.
However, for making use of the real-bogus classifier during the upcoming sprint, I think we can still use it in its current hacky, machine-dependent form.

As for whether it is instrument- or data-specific, I lean towards the former. It should NOT depend on the test data -- but we will have various versions, and the user should be able to specify them when defining the pipeline, which is another discussion to happen in the above ticket I guess.

kfindeisen: If the model is instrument-wide then it probably doesn't belong in a repository; I've said as much on DM-36742 (thanks for pointing me to it!). Putting it in one of the obs packages would also solve the present issue of having a platform-dependent path, since then the path would be relative to a package directory (which is required to be set up by the dataset definition).

However, I disagree that this is an acceptable hack in the meantime, because it will break our nightly builds indefinitely. I'd recommend making this issue blocked on DM-36742 and using a local pipeline for the sprint, or at least putting the `TransiNetTask` pipeline into a separate pipeline that isn't in any of the standard execution paths (i.e., not `<dataset>/ApVerify.yaml` or `<dataset>/ApVerifyWithFakes.yaml`, nor any of the pipelines they include).

pipelines/LSSTCam-imSim/ApVerify.yaml
kfindeisen: I'm flattered, but it's not standard practice in DM to cite people for pointing out corrections on the PR. Just [write the commit as if you'd done it that way the first time](https://developer.lsst.io/work/flow.html#rebase-commits-from-code-reviews-rather-than-having-review-feedback-commits).

NimSed:
@kfindeisen, the model package is now stored in the neighboring package (rbClassify_data). Here's the relevant PR:
https://github.com/lsst-dm/rbClassifier_data/pull/1

Although I still don't think the problem with the nightly builds is fully resolved, according to the discussions and decisions under https://jira.lsstcorp.org/browse/DM-37478 this is the solution we want to go ahead with (and it is less hacky at least).

Please let me know if there are any more points.

pipelines/LSSTCam-imSim/ApVerify.yaml
```diff
@@ -15,6 +15,11 @@ tasks:
     class: lsst.ap.association.DiaPipelineTask
     config:
       doPackageAlerts: True
+  rbClassify:
+    class: lsst.meas.transiNet.TransiNetTask
+    config:
+      modelPackageName: 'neighbor:///rbResnet50-DC2'
```
kfindeisen, line 21: Note that, in order for this to work on Jenkins, we'd need to modify https://github.com/lsst-dm/jenkins-dm-jobs/blob/main/pipelines/scipipe/ap_verify.groovy to download the neighbor repository in addition to the `ap_verify` dataset.

kfindeisen (commented):
I believe that merging this will still break the nightly `ap_verify` build, because we don't have the supporting infrastructure in place.

(Also, note that @isullivan is the reviewer of record -- I'm just heckling, so my opinion doesn't count 😉)

NimSed:
@kfindeisen please check and see whether the issue you've mentioned [here](https://jira.lsstcorp.org/browse/DM-36742?focusedCommentId=2467235&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2467235) is resolved.
Did you catch it in the usual `pytest` tests, or when running it through jenkins?

pipelines/LSSTCam-imSim/ApVerify.yaml
NimSed: Given the implementation of model packages and the current git-repository-based backend, I see this discussion as resolved. Please feel free to reopen it in case I'm mistaken.

pipelines/LSSTCam-imSim/ApVerifyWithFakes.yaml
```diff
@@ -11,3 +11,12 @@ imports:
...
   # because the changes made by that file and ../ApVerifyWithFakes.yaml are
   # hard to separate.
+
+tasks:
+  rbClassify:
+    class: lsst.meas.transiNet.TransiNetTask
+    config:
+      modelFile: testModel
+      weightsFile: '/home/nima/meas_transiNet/tests/data/checkpoint.pth.zip'
+      connections.coaddName: 'goodSeeing'
```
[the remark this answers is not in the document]
NimSed: Given the implementation of model packages and the current git-repository-based backend, I see this discussion as resolved. Please feel free to reopen it in case I'm mistaken.

kfindeisen:
Yes, that looks like the current `TransiNetConfig` API. No, I did not try to run any tests -- what would have been the point?

PR #176, opened 2023-01-10
DM-35207: Reset visitSummary connections after upstream DRP-focused changes.

TallJimbo:


pipelines/Conversions.yaml
```diff
@@ -32,6 +32,10 @@ tasks:
   # Conversion of visitSummary [visit-level afw.table] to visitTable [instrument-level Parquet]
   makeVisitTable:
     class: lsst.pipe.tasks.postprocess.MakeVisitTableTask
+    config:
+      connections.visitSummaries: visitSummary
```
kfindeisen, line 36: If the goal is to restore the connection name to what it was before lsst/pipe_tasks#741, then please make this `{calexpType}visitSummary`. We get a lot of bugs from importing pipelines that have connections overridden to fixed values; it's better that it always has the templates we expect it to, even if we don't expect to use them.

TallJimbo: Will do; looks like that will also let me drop these overrides from the `ConversionsForFakes.yaml` since they'll be imported.

kfindeisen: But `consolidateVisitSummaryWithFakes` is not imported, it's brand-new.

TallJimbo: Ah, thanks :man_facepalming: .

TallJimbo: Actually, I think that's okay, I think, because `makeVisitTable` and `makeCcdVisitTable` are not new there, and they're the ones getting the overrides.  `ConsolidateVisitSummaryTask`'s defaults have not been changed upstream, only the defaults of the tasks that previously consumed `{calexpType}visitSummary`.

laurenam:
@kfindeisen, could I ask you to approve this one once you’re satisfied?

kfindeisen (approved):
So long as the config overrides at https://github.com/lsst/ap_verify/blob/c560b20640ab36416a898e2fe435d5766c0b9676/pipelines/ConversionsForFakes.yaml#L45-L54 are appended to those in `Conversions.yaml` instead of overwriting them (I always forget when that happens), I think this is fine.

PR #183, opened 2023-04-18
DM-31924: Update mock tasks to use configurable ID generators.

TallJimbo:
This reflects updates to the mocked tasks.

### DIFF
--- a/python/lsst/ap/verify/testPipeline.py
+++ b/python/lsst/ap/verify/testPipeline.py
@@ -34,7 +34,6 @@ import lsst.geom as geom
 import lsst.afw.image as afwImage
 import lsst.afw.math as afwMath
 import lsst.afw.table as afwTable
-import lsst.obs.base as obsBase
 from lsst.pipe.base import PipelineTask, Struct
 from lsst.ip.isr import IsrTaskConfig
 from lsst.ip.diffim import GetTemplateConfig, AlardLuptonSubtractConfig, DetectAndMeasureConfig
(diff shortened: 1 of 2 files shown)


python/lsst/ap/verify/testPipeline.py
```diff
@@ -149,10 +148,20 @@ def run(self, ccdExposure, *, camera=None, bias=None, linearizer=None,
...
 
 
+class MockCharacterizeImageConfig(CharacterizeImageConfig):
+
+    def setDefaults(self):
+        super().setDefaults()
+        # Avoid delegating to lsst.obs.base.Instrument specialization for the
+        # data ID packing algorithm to use, since test code often does not use a
+        # real Instrument in its data IDs.
+        self.idGenerator.packer.name = "observation"
```
kfindeisen, line 158: I'm confused -- why can't you just use `CharacterizeImageConfig` as-is?

Since [some of the tests assume that the correct connections are drawn from the "real" tasks](https://github.com/lsst/ap_verify/blob/main/tests/test_testPipeline.py#L48-L49), creating custom configs is a dangerous precedent even if they don't fiddle with the connections themselves.

python/lsst/ap/verify/testPipeline.py
```diff
@@ -463,17 +505,28 @@ def run(self, diaSourceCat, diffIm, band, ccdVisitId, funcs=None):
...
 
 
+class MockDiaPipelineConfig(DiaPipelineConfig):
+
+    def setDefaults(self):
+        super().setDefaults()
+        # Avoid delegating to lsst.obs.base.Instrument specialization for the
+        # data ID packing algorithm to use, since test code often does not use a
+        # real Instrument in its data IDs.
+        self.idGenerator.packer.name = "observation"
```
kfindeisen, line 515: What exactly is this ID generator, that _every_ task needs to use it? It seems like the sort of thing that `DiaPipelineTask` wouldn't need to bother with.

TallJimbo: It's a way to pack data IDs (`{visit, detector}` ones, in this case) and maybe release IDs into single integers.  It's going everywhere `lsst.obs.base.ExposureIdInfo` was once used.

python/lsst/ap/verify/testPipeline.py
kfindeisen: That might be a good idea. The test suite that actually _uses_ `MockApPipe.yaml`, `test_driver.py`, is based on [ap_verify_testdata](https://github.com/lsst/ap_verify_testdata), which uses `LSSTCam-imSim` as its instrument.

timj: `obs_base` has a `Instrument` specifically designed for testing if that helps at all.

parejkoj (commented):
Echoing Krzysztof's comments.

python/lsst/ap/verify/testPipeline.py
TallJimbo: Ah, it turns out `test_driver.py` was already fine and it was `test_testPipeline.py` that was all that was giving me trouble.  And @timj 's suggestion of `lsst.obse.base.instrument_tests.DummyCam` is the perfect solution for that.

PR #186, opened 2023-05-26
DM-39214: Move ingredient pipelines into _ingredients dir

leeskelvin:


pipelines/README.md
```diff
@@ -0,0 +1,15 @@
+# Pipeline Definitions
+
+This directory contains pipeline definition YAML files which are used when processing data with the LSST Science Pipelines.
+
+The pipelines defined here come in three flavors: camera-specific (within named directories), camera-agnostic (top-level, if any), and building-block ingredients (within the [\_ingredients](_ingredients) directory).
+Pipelines within the ingredients directory tend to be imported by other pipelines, and are not intended to be used directly by end-users for science purposes.
```
kfindeisen, line 6: The original description sounds a bit odd, as if you were documenting a natural phenomenon:
```suggestion
Pipelines within the ingredients directory are meant to be imported by other pipelines, and are not intended to be used directly by end-users for science purposes.
```

pipelines/_ingredients/ApVerify.yaml
```diff
@@ -3,10 +3,10 @@
...
 description: Fully instrumented AP pipeline
 imports:
-  - location: $AP_PIPE_DIR/pipelines/ApPipe.yaml
-  - location: $AP_VERIFY_DIR/pipelines/MetricsRuntime.yaml
-  - location: $AP_VERIFY_DIR/pipelines/MetricsMisc.yaml
-  - location: $AP_VERIFY_DIR/pipelines/Conversions.yaml
+  - location: $AP_PIPE_DIR/pipelines/_ingredients/ApPipe.yaml
+  - location: $AP_VERIFY_DIR/pipelines/_ingredients/MetricsRuntime.yaml
+  - location: $AP_VERIFY_DIR/pipelines/_ingredients/MetricsMisc.yaml
+  - location: $AP_VERIFY_DIR/pipelines/_ingredients/Conversions.yaml
```
kfindeisen, line 9: The "not meant to be run" status of the metrics pipelines is a bit messy. It's true that they're designed to be used as part of `ApVerify*.yaml`, mostly because they depend on external `parameter` definitions, but they're also (supposed to be) frequently updated as developers add new metrics for AP testing. I'm worried that this second usage will be harmed by hiding them in `_ingredients`.

pipelines/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+The `pipetask build` command can be used to expand a pipeline YAML and resolve any imports for the purposes of visualizing it.
+For example, to visualize the `processCcd` subset from the [LSSTCam-imSim ApVerify pipeline](https://github.com/lsst/ap_verify/blob/main/pipelines/LSSTCam-imSim/ApVerify.yaml) pipeline, run:
```
kfindeisen, line 9: ImSim is the colloquial name of the instrument:
```suggestion
For example, to visualize the `processCcd` subset from the [ImSim ApVerify pipeline](https://github.com/lsst/ap_verify/blob/main/pipelines/LSSTCam-imSim/ApVerify.yaml) pipeline, run:
```

pipelines/README.md
kfindeisen, line 6: Nothing in `ap_verify` should be used for science purposes, something I occasionally butt heads with would-be users about. Perhaps: 
```suggestion
Pipelines within the ingredients directory tend to be imported by other pipelines, and are not intended to be used directly by end-users for verification purposes.
```

pipelines/_ingredients/README.md
```diff
@@ -0,0 +1,8 @@
+# Pipeline Definition Ingredients
+
+This directory contains ingredient pipeline definition YAML files.
+
+The pipelines defined here are intended to be used as building blocks for more complex pipelines.
+As such, they are not intended to be used directly by end-users for science purposes.
+
+Science ready pipeline definitions are located in the [pipelines](..) directory.
```
kfindeisen, line 8:
```suggestion
The pipelines defined here are intended to be used as building blocks for more complex pipelines.
As such, they are not intended to be used directly by end-users for verification purposes.

Standalone pipeline definitions are located in the [pipelines](..) directory.
```

kfindeisen (approved):
I'm a bit concerned about applying the boilerplate to some of `ap_verify`'s weirdness, but otherwise looks good.

pipelines/_ingredients/ApVerify.yaml
leeskelvin: It is *not* intended that developers should shy away from making any commits to pipeline YAMLs within the `_ingredients` directory. Rather, it is meant to help guide end-users towards a camera-specific implementation of a pipeline rather than a generic one.

leeskelvin: If these metrics pipelines *are* intended to be directly used by end-users, then we should find another home for them. Either in a camera-specific implementation, or as a camera-agnostic implementation in the top-level `pipelines` directory.

pipelines/README.md
leeskelvin: As on the other PR, I think it may be nicer to label this `LSSTCam-imSim` in line with [DM-36796](https://jira.lsstcorp.org/browse/DM-36796). Either way, we should still refer to `LSSTCam-imSim` at some point to help discoverability of these pipeline definitions outside of clicking this link.

pipelines/README.md
leeskelvin: This may be a matter of definition - isn't verification a part of the scientific process? If this is a point of contention then we could just remove the latter part of this sentence, ending with "by end-users."

pipelines/_ingredients/ApVerify.yaml
kfindeisen: Ok, let's leave them where they are.

pipelines/README.md
kfindeisen: When I hear a phrase like "science runs" or "scientific processing", it usually means running a pipeline on a generic set of data. The `ap_verify` pipelines, unlike the `ap_pipe` pipelines, are not designed to be run on anything other than a packaged `ap_verify` dataset -- it might work, but it's very Not Supported.

I'm fine with just removing the final phrase.

pipelines/_ingredients/README.md
leeskelvin: Modifying this sentence to end with "by end-users.", as discussed above.

leeskelvin: Replacing "Standalone" with "User-facing". The term standalone to me implies that these may be used in isolation, but more often than not they import a series of other pipeline YAMLs in order to function correctly.

PR #187, opened 2023-05-28
DM-39378: Rename "spuriousness" to "reliability" in the DiaSource table

rai-harshit:


parejkoj (approved):

parejkoj:
After discussion, we'll merge this at the same time as DM-37196, to keep things consistent.

PR #189, opened 2023-06-08
DM-39605: Use butler.dimensions rather than butler.registry.dimensions

timj:
The latter will be deprecated in the future.


kfindeisen (approved):
I don't think "will be deprecated in the future" is reasonable grounds for removal, but I have no objections to the change itself.

timj:
We want people to use `butler.dimensions` because in a few months that will be the only way to do it. I'm getting ahead of things. There is no active deprecation for `butler.registry.dimensions` at the moment because sometimes people only have a `Registry` object but if they are starting from butler we want them to do it this way and there is no harm in doing it this way now.

PR #196, opened 2023-08-03
DM-40210: Clean up ap_pipe and ap_verify pipelines

kfindeisen:
This PR merges the `rbClassify` task from instrument-specific pipelines to the base (`_ingredients`) pipeline and removes some configs that are redundant at the instrument level. This leaves the pipelines cleaner and with clearer responsibilities, and will prevent changes to the base pipeline from unexpectedly not working.

Unfortunately, the `ApVerify` pipeline still contains a few elements that should be in `ApPipe` but that shouldn't be moved at this time. These have been marked with TODO comments.


rai-harshit (approved):
Went through the Jira ticket and reviewed the code. Looks good and in alignment with ticket description/comments.

PR #197, opened 2023-08-22
DM-40470: Add exclusions for metadata datasets to ap_verify pipelines

parejkoj:


pipelines/_ingredients/ApVerify.yaml
```diff
@@ -5,6 +5,9 @@ description: Fully instrumented AP pipeline
 imports:
   - location: $AP_PIPE_DIR/pipelines/_ingredients/ApPipe.yaml
   - location: $AP_VERIFY_DIR/pipelines/_ingredients/MetricsRuntime.yaml
+    exclude:
+      - timing_calibrateImage
+      - cputiming_calibrateImage
```
kfindeisen, line 10: I'd recommend adding the analogous lines to `ApVerifyWithFakes.yaml` and `ApVerifyCalibrateImage.yaml`, to prevent similar problems when we try to run them (or do they run as-is?).

kfindeisen (changes requested):
Can you confirm that all three pipelines run now? I'd be surprised if it were exclusive to `ApVerify.yaml`.

pipelines/_ingredients/ApVerify.yaml
parejkoj: Ah, I forgot to check the fakes version: that needed the fix as well. I already had the exclusions in the new pipeline.

kfindeisen (approved):

PR #198, opened 2023-08-23
DM-40110: Switch from Gaia DR2 to DR3

cmsaunders:


parejkoj (commented):
Did you also add dr3 files to the `ap_verify_ci_*` test data sets?

PR #203, opened 2023-09-15
DM-33453: Update pipeline references in response to RFC-775.

erinleighh:
Changed `HyperSuprimeCam` to `HSC` and `DarkEnergyCamera` to `DECam` in response to RFC-775.


kfindeisen (approved):
If you're going to keep the symlinks, please separate them into their own commit and file a ticket for their eventual removal -- otherwise we'll be stuck dealing with the extra friction indefinitely.

My usual concern about the commit text still applies.

kfindeisen (approved):

PR #204, opened 2023-09-20
DM-39796: Do not fail if doWriteMatchesDenormalized=True

parejkoj:


kfindeisen (approved):
Can you please add an explanation (e.g., to the commit message) about why the old value failed? The dummy pipeline really shouldn't care about anything other than having outputs of the right type.

parejkoj:
I updated the message: the problem was exactly that the output wasn't the right type (and we never checked this, because in these tests we never tried to output the denormalized match list).

PR #205, opened 2023-10-27
DM-41024: Add longTrailedSources to testPipeline unit tests

bsmartradio:
longTrailedSources needs to be added to the unit test output otherwise ap_verify fails to scons.

### DIFF
--- a/tests/test_testPipeline.py
+++ b/tests/test_testPipeline.py
@@ -120,7 +120,8 @@ class MockTaskTestSuite(unittest.TestCase):
         butlerTests.addDatasetType(cls.repo, "deepDiff_diaSrcTable", cls.visitId.keys(), "DataFrame")
         butlerTests.addDatasetType(cls.repo, "visitSsObjects", cls.visitOnlyId.keys(), "DataFrame")
         butlerTests.addDatasetType(cls.repo, "apdb_marker", cls.visitId.keys(), "Config")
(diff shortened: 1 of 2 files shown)


tests/test_testPipeline.py
```diff
@@ -121,6 +121,7 @@ def setUpClass(cls):
         butlerTests.addDatasetType(cls.repo, "visitSsObjects", cls.visitOnlyId.keys(), "DataFrame")
         butlerTests.addDatasetType(cls.repo, "apdb_marker", cls.visitId.keys(), "Config")
         butlerTests.addDatasetType(cls.repo, "deepDiff_associDiaSrc", cls.visitId.keys(), "DataFrame")
+        butlerTests.addDatasetType(cls.repo, "deepDiff_longTrailedSrc", cls.visitId.keys(), "DataFrame")
```
mrawls, line 124: (If you decide to go with `longTrailedSources` instead of `Src`, you'll need to also change it here)

mrawls: WAIT, is the line above this a typo? It says `associDiaSrc` and I am 99% sure it should be `assocDiaSrc`, what is happening here?

mrawls (approved):
Your changes are fine, but I'm surprised everything has been fine with a typo in `assocDiaSrc` prior to now, in the test. If it's not too much trouble to fix that and verify (ha) it doesn't break anything, that would be appreciated.

PR #213, opened 2024-02-15
DM-42419: Break up ApVerify into pure with- and without-fakes pipelines

kfindeisen:
This PR removes all unnecessary without-fakes processing from `ApVerifyWithFakes`, and moves all metrics that are better measured without fakes to `ApVerify`. With no more need to disambiguate, it also removes the `WithFakes` suffix from individual tasks in `ApVerifyWithFakes`.

****

- [X] Is the Sphinx documentation up-to-date?


BrunoSanchez (commented):
Looks perfect. 
Maybe the comments for clarifying would be useful for future, but probably not needed.

BrunoSanchez (approved):
Looks perfect. 
Maybe the comments for clarifying would be useful for future, but probably not needed.

PR #216, opened 2024-03-07
DM-40400: Create total AP pipeline timing metric

kfindeisen:
This PR adds the timing metric introduced on lsst/ap_pipe#168 to the standard runtime metrics for `ApVerify`. Like all timing metrics, it is not run on `ApVerifyWithFakes`.

****

### DIFF
--- a/pipelines/_ingredients/MetricsRuntime.yaml
+++ b/pipelines/_ingredients/MetricsRuntime.yaml
@@ -267,3 +267,12 @@ tasks:
       connections.metric: PackageAlertsMemory
       connections.labelName: diaPipe
       target: diaPipe:alertPackager.run
+  timing_apPipe:
+    class: lsst.ap.pipe.metrics.PipelineTimingMetricTask
+    config:
+      connections.package: ap_pipe
+      connections.metric: ApPipelineTime
+      connections.labelStart: isr
+      connections.labelEnd: diaPipe
+      targetStart: isr.run
+      targetEnd: diaPipe.run


pipelines/_ingredients/MetricsRuntime.yaml
```diff
@@ -267,3 +267,12 @@ tasks:
       connections.metric: PackageAlertsMemory
       connections.labelName: diaPipe
       target: diaPipe:alertPackager.run
+  timing_apPipe:
+    class: lsst.ap.pipe.metrics.PipelineTimingMetricTask
+    config:
+      connections.package: ap_pipe
+      connections.metric: ApPipelineTime
+      connections.labelStart: isr
+      connections.labelEnd: diaPipe
```
parejkoj, line 276: Maybe worth a comment here that diaPipe is what sends the alerts, and thus is the endpoint of our runtime requirement? That might not be obvious to someone reading, especially if our pipeline grows some post-diaPipe afterburners.

We may also implement some things in diaPipe after packageAlerts (writing `detectorVisitProcessingSummary` comes to mind) so it's always going to be an upper-bound on the runtime.

parejkoj (approved):
One suggestion, otherwise this makes the use of the new metric quite clear.

pipelines/_ingredients/MetricsRuntime.yaml
kfindeisen: To clarify, this metric is pretty much irrelevant for "our runtime requirement", because the batch processing environment is completely different from the prompt one. Based on my experiments with the three `ap_verify` datasets, a good chunk of the time measured by this metric is time spent running _other data IDs_ tasks.

PR #222, opened 2024-04-18
DM-41063: Remove trailedSourceFilter and add to filterDiaSourceCatalog

bsmartradio:
Remove longTrailedSources from `diaPipe`, change its output to `ArrowAstropy`, and add longTrailedSources to output from `filteredDiaSourceCatalog`
****

### DIFF
--- a/python/lsst/ap/verify/testPipeline.py
+++ b/python/lsst/ap/verify/testPipeline.py
@@ -442,22 +442,33 @@ class MockFilterDiaSourceCatalogTask(PipelineTask):
     ConfigClass = FilterDiaSourceCatalogConfig
     _DefaultName = "notFilterDiaSourceCatalog"
 
-    def run(self, diaSourceCat):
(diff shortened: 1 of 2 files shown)


python/lsst/ap/verify/testPipeline.py
```diff
@@ -455,8 +455,10 @@ def run(self, diaSourceCat):
         results : `lsst.pipe.base.Struct`
             Results struct with components.
         """
+        # TODO Add docstrings for diffIm
```
mrawls, line 458: Good idea 😜

mrawls (approved):
Approving with the assumption you will resolve the docstring TODO.

python/lsst/ap/verify/testPipeline.py
bsmartradio: Added docstrings and updated the results docstrings to be more comprehensive.

PR #210, opened 2024-01-29
DM-40798: Exclude sky sources from RB classification

rai-harshit:
{Summary of changes. Prefix PR title with JIRA issue.}

****

- [x] Do unit tests pass (`scons` and/or `stack-os-matrix`)?
- [x] Did you run `ap_verify.py` on at least one of [the standard datasets](https://pipelines.lsst.io/v/daily/modules/lsst.ap.verify/datasets.html#supported-datasets)?
      For changes to metrics, the `print_metricvalues` script from `lsst.verify` will be useful.
- [x] Is the Sphinx documentation up-to-date?

PR #224, opened 2024-04-26
DM-43416: Migrate AP code to external APDB configs

kfindeisen:
This PR modifies the `ap_verify` framework and the metrics pipelines to use new-style configs instead of handling `ApdbConfig` objects directly. It must be merged after lsst/ap_association#210 and lsst/verify#124.

****

### DIFF
--- a/python/lsst/ap/verify/pipeline_driver.py
+++ b/python/lsst/ap/verify/pipeline_driver.py
@@ -36,8 +36,7 @@ import subprocess
 import logging
 
 import lsst.ctrl.mpexec.execFixupDataId  # not part of lsst.ctrl.mpexec
-import lsst.ctrl.mpexec.cli.pipetask
-from lsst.ap.pipe.make_apdb import makeApdb
+import lsst.dax.apdb as daxApdb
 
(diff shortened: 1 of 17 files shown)


python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -35,9 +35,9 @@
 import subprocess
 import logging
 
+import lsst.dax.apdb as daxApdb
```
isullivan, line 38: Minor, but should this be after `ctrl.mpexec` imports?

isullivan (approved):
Looks good, very minor comment.

python/lsst/ap/verify/pipeline_driver.py
kfindeisen: I'm pretty sure `dax.apdb` is much earlier in the dependency tree than `ctrl.mpexec`...

kfindeisen: Hmm, [it's actually undefined](https://peps.python.org/pep-0008/#imports). Fine.

PR #228, opened 2024-07-16
DM-42576

Gerenjie:
Adding mpSkyEphemerisQuery, which replaces SkyBot in querying mpSky for ssObjects in the pre-load step.

ap_verify.py on ap_verify_ci_cosmos_pdr2 is failing with `lsst.daf.butler._exceptions.MissingDatasetTypeError: "Dataset type(s) ['RegionTimeInfo'] are not registered."`


timj:
If a task uses a newly-defined dataset type then pipetask needs to be told that it can create new dataset types. We don't create by default because sometimes there were typos in connections and we ended up with lots of incorrect dataset types in the registry.

PR #229, opened 2024-07-26
DM-42576: Write task to compute predicted positions for an upcoming visit

Gerenjie:

****

- [x ] Do unit tests pass (`scons` and/or `stack-os-matrix`)?
- [x ] Did you run `ap_verify.py` on at least one of [the standard datasets](https://pipelines.lsst.io/v/daily/modules/lsst.ap.verify/datasets.html#supported-datasets)?
- [ x] Is the Sphinx documentation up-to-date?

### DIFF
--- a/doc/lsst.ap.verify/limitations.rst
+++ b/doc/lsst.ap.verify/limitations.rst
@@ -1,0 +1,14 @@
+.. py:currentmodule:: lsst.ap.verify
+
+.. _ap-verify-request:
+
+#################
+Known Limitations 
+#################
+
+mpSkyEphemerisQueryTask
+=======================
+
+The AP pipelines now run :py:class:`lsst.ap.association.MPSkyEphemerisQueryTask`, which includes a request to mpSky.
+mpSky is not expected to host ephemerides for all test datasets, (or be reachable from Jenkins) so it will often (or always) fail.
+When requests fail, mpSkyEphemerisQuery will return nothing, and diaPipe will skip ssoAssociation.
--- a/doc/lsst.ap.verify/index.rst
+++ b/doc/lsst.ap.verify/index.rst
@@ -25,6 +25,7 @@ Using lsst.ap.verify
    datasets
    new-metrics
    command-line-reference
+   limitations
 
 .. _lsst.ap.verify-contributing:
 


kfindeisen:
It looks like something went wrong with the rebase -- AFAICT "Added logic for MPSkyEphemerisQueryTask" should be the only commit on the PR. I'll continue the review on just 9ae80caf7ce57e4626414b029e7685cca19aa673, but you might want to re-rebase on your end in the meantime.

tests/test_testPipeline.py
```diff
@@ -351,14 +379,14 @@ def testMockDiaPipelineTask(self):
         pipelineTests.assertValidOutput(task, result)
 
         self.butler.put(pandas.DataFrame(), "deepDiff_diaSrcTable", self.visitId)
-        self.butler.put(pandas.DataFrame(), "visitSsObjects", self.visitId)
+        self.butler.put(pandas.DataFrame(), "preloaded_ssObjects", self.groupId)
```
kfindeisen, line 382: It's a bit strange that you can drop `visitSsObjects`, because the default `DiaPipelineConnections` still takes them as inputs. That and the fix below means this code wasn't being exercised in the first place. 😞 

I'd recommend leaving this with `visitSsObjects`, since what we're supposed to be testing here is whether `MockDiaPipelineTask` is still representative of `DiaPipelineTask`, which has not been modified to support the new group-based dataset.

kfindeisen (approved):
Looks good, mostly style comments. Please remember to [describe the PR in the title](https://developer.lsst.io/work/flow.html#make-a-pull-request) -- people also come across PRs from the GitHub side, and most can't remember arbitrary 5-digit codes. 🙂

tests/test_testPipeline.py
Gerenjie: tests/test_testPipeline.py::MockTaskTestSuite::testMockDiaPipelineTask fails when this is left with `visitSsObjects` but does not when this is left with `preloaded_SsObjects` (new capitalization).

kfindeisen:
Before I try to re-review this, can you please re-rebase onto `main` to drop everybody else's commits? Either @isullivan or I can help you with that if you need it.

Gerenjie:
> Before I try to re-review this, can you please re-rebase onto `main` to drop everybody else's commits? Either @isullivan or I can help you with that if you need it.

Done!

doc/lsst.ap.verify/limitations.rst
```diff
@@ -0,0 +1,12 @@
+.. py:currentmodule:: lsst.ap.verify
+
+.. _ap-verify-request:
+
+########################################
+Known Limitations 
+########################################
```
kfindeisen, line 7: Sphinx requires that the header symbol lines be exactly as long as the header text. Please do run `package-docs build`, it will catch such things.

doc/lsst.ap.verify/limitations.rst
```diff
@@ -0,0 +1,12 @@
...
+
+mpSkyEphemerisQueryTask
+=======================
+
+The AP pipelines now run :py:class:`lsst.ap.association.MPSkyEphemerisQueryTask`, which includes a request to mpSky.  mpSky is not expected to host ephemerides for all test datasets, so it will often (or always) fail. When requests fail, mpSkyEphemerisQuery should return an empty DataFrame, and no detections will be associated.
```
kfindeisen, line 12: Please [write each sentence on a new line](https://developer.lsst.io/restructuredtext/style.html#text-wrapping); this makes it easier to spot future changes (as opposed to, say, changing one word and `git diff` counting it as replacing the whole paragraph).

kfindeisen: The last sentence is out of date now. 🙂

kfindeisen (approved):
Looks good, just a few comments about the new docs.

kfindeisen:
Please squash the "Formatting" commit, as it's a classic "fixup". You'll also need to rebase again; fortunately it's a simple resolution because `testPipeline.py` and `test_testPipeline.py` can just be deleted.

PR #230, opened 2024-08-03
DM-31833: Move LoadDiaCatalogs to standalone pipelineTask 

isullivan:

****

- [ x] Is the Sphinx documentation up-to-date?

### DIFF
--- a/python/lsst/ap/verify/testPipeline.py
+++ b/python/lsst/ap/verify/testPipeline.py
@@ -34,7 +34,7 @@ import lsst.geom as geom
 import lsst.afw.image as afwImage
 import lsst.afw.math as afwMath
 import lsst.afw.table as afwTable
-from lsst.ap.association import (TransformDiaSourceCatalogConfig,
+from lsst.ap.association import (LoadDiaCatalogsConfig, TransformDiaSourceCatalogConfig,
                                  DiaPipelineConfig, FilterDiaSourceCatalogConfig)
 from lsst.pipe.base import PipelineTask, Struct
 from lsst.ip.isr import IsrTaskConfig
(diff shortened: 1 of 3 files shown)


python/lsst/ap/verify/testPipeline.py
kfindeisen, line 1: Please add `MockLoadDiaCatalogsTask` to `tests/MockApPipe.yaml`; otherwise there's no point in having it.

tests/test_testPipeline.py
```diff
@@ -39,7 +39,10 @@
     MockCalibrateTask, MockGetTemplateTask, \
     MockAlardLuptonSubtractTask, MockDetectAndMeasureTask, MockTransformDiaSourceCatalogTask, \
     MockRBTransiNetTask, MockDiaPipelineTask, MockFilterDiaSourceCatalogTask, \
-    MockSpatiallySampledMetricsTask
+    MockSpatiallySampledMetricsTask, MockLoadDiaCatalogsTask
+from lsst.daf.butler import Timespan
+from lsst.pipe.base.utils import RegionTimeInfo
+from lsst.sphgeom import UnitVector3d
```
kfindeisen, line 45: I recommend sticking to the established (dependency) order, so that it's easier for future developers to see that the import already exists.

tests/test_testPipeline.py
```diff
@@ -141,6 +153,27 @@ def setUp(self):
         super().setUp()
         self.butler = butlerTests.makeTestCollection(self.repo, uniqueId=self.id())
 
+    def testMockLoadDiaCatalogsTask(self):
+        config = MockLoadDiaCatalogsTask.ConfigClass()
+        config.apdb_config_url = "testing_only"
+        task = MockLoadDiaCatalogsTask(config=config)
+        pipelineTests.assertValidInitOutput(task)
+        points = [UnitVector3d.Z(), UnitVector3d.X(), UnitVector3d.Y()]
+        region = lsst.sphgeom.ConvexPolygon(points)
```
kfindeisen, line 162: If you're making a dummy input, `sphgeom.Circle` has [a default constructor](http://doxygen.lsst.codes/stack/doxygen/x_mainDoxyDoc/classlsst_1_1sphgeom_1_1_circle.html#acc3ef9b79e6d97e854f9ee1eed81c42c) so you don't have to import `UnitVector3d`. You get an empty region either way.

tests/test_testPipeline.py
```diff
@@ -347,9 +380,13 @@ def testMockDiaPipelineTask(self):
         task = MockDiaPipelineTask(config=config)
         pipelineTests.assertValidInitOutput(task)
         result = task.run(pandas.DataFrame(), pandas.DataFrame(), afwImage.ExposureF(),
-                          afwImage.ExposureF(), afwImage.ExposureF(), 42, 'k')
+                          afwImage.ExposureF(), afwImage.ExposureF(), pandas.DataFrame(),
+                          pandas.DataFrame(), pandas.DataFrame(), 'k', 42)
```
kfindeisen, line 384: I think the `DiaPipelineTask` API has gotten to the point where we should be using keywords. 😬

kfindeisen: [Same question](https://github.com/lsst/ap_association/pull/230/files#r1706125095) about why this wasn't caught earlier. `MockDiaPipelineTask` allows `idGenerator=None`, but [the original does not](https://github.com/lsst/ap_association/blob/main/python/lsst/ap/association/diaPipe.py#L392).

(And again, such a fix should have been a separate commit.)

kfindeisen (approved):
Looks good, just style comments.

tests/test_testPipeline.py
isullivan: I'll clean up some of the mocks, but it looks like they were not updated when we removed `ccdExposureIdBits`. This would seem to be an argument in favor of removing `test_testPipeline`, since it doesn't necessarily catch changes to Task API's unless the Mocks in `testPipeline.py` are also updated.

PR #233, opened 2024-08-23
DM-45959: Update timing metrics for LoadDiaCatalogsTask

isullivan:
{Summary of changes. Prefix PR title with JIRA issue.}

****

- [ x] Do unit tests pass (`scons` and/or `stack-os-matrix`)?
- [x ] Did you run `ap_verify.py` on at least one of [the standard datasets](https://pipelines.lsst.io/v/daily/modules/lsst.ap.verify/datasets.html#supported-datasets)?
      For changes to metrics, the `print_metricvalues` script from `lsst.verify` will be useful.
- [x ] Is the Sphinx documentation up-to-date?

PR #234, opened 2024-08-26
DM-45959: Fix incorrect metadata dimensions for *timing_loadDiaCatalogs.

kfindeisen:
This PR fixes a bug introduced in #233 that causes the timing metrics to try to load metadata of the wrong dimensions, causing quantum graph failure.

****


isullivan (approved):
Thank you for fixing this, I didn't realize the dimensions needed to be changed here from the base class.

PR #236, opened 2024-09-17
DM-43077: Convert all tasks to use CalibrateImageTask outputs

parejkoj:
Remove now unused datasets.

****

- [X] Is the Sphinx documentation up-to-date?


pipelines/_ingredients/ConversionsForFakes.yaml
```diff
@@ -35,7 +35,6 @@ tasks:
     class: lsst.pipe.tasks.postprocess.ConsolidateVisitSummaryTask
     config:
       connections.calexpType: parameters.fakesType
-      connections.calexp: "{calexpType}initial_pvi"
```
kfindeisen, line 38: I'm pretty sure this one won't work, because the default is `initial_pvi` without referencing `calexpType`. Can you confirm that the fakes pipeline has the correct structure?

parejkoj: Good question. `calibrateImage` isn't setup to work with the fakes pipeline as it is used in AP, I don't think. This may require some discussion: I put it on the schedule for Monday's meeting.

parejkoj: On second thought, maybe it's fine: ci_ap passed on Jenkins, and that runs the fakes pipeline for DC2.

kfindeisen: Just for the record, this question cross-references lsst/pipe_tasks#979.

kfindeisen: I've confirmed that the pipeline is incorrect; the task converts `initial_pvi` into `fakes_visitSummary`. Please fix (and in the future, please do this check yourself).

ApVerify

kfindeisen (changes requested):
The PR changes the pipeline structure by replacing a fakes input with a non-fakes input.

kfindeisen (approved):

PR #239, opened 2024-10-14
DM-46599: Stop using Butler regular expression searches

dhirving:
Using regular expressions to search the Butler for collection names is deprecated in RFC-1040.

****

- [x] Do unit tests pass (`scons` and/or `stack-os-matrix`)?
- [x] Did you run `ap_verify.py` on at least one of [the standard datasets](https://pipelines.lsst.io/v/daily/modules/lsst.ap.verify/datasets.html#supported-datasets)?

### DIFF
--- a/python/lsst/ap/verify/pipeline_driver.py
+++ b/python/lsst/ap/verify/pipeline_driver.py
@@ -277,7 +277,10 @@ def _getCollectionArguments(workspace, reuse):
     # skip-existing-in would work around that, but would lead to a worse bug in
     # the case that the user is alternating runs with and without --clean-run.
     # registry.refresh()
-    oldRuns = list(registry.queryCollections(re.compile(workspace.outputName + r"/\d+T\d+Z")))
+    collectionPattern = re.compile(workspace.outputName + r"/\d+T\d+Z")
+    oldRuns = list(registry.queryCollections(workspace.outputName + "/*"))
+    oldRuns = [run for run in oldRuns if collectionPattern.fullmatch(run)]
+
     if reuse and oldRuns:
         args.extend(["--extend-run", "--skip-existing"])
     return args
--- a/python/lsst/ap/verify/workspace.py
+++ b/python/lsst/ap/verify/workspace.py
@@ -26,7 +26,6 @@ __all__ = ["Workspace", "WorkspaceGen3"]
 import abc
 import os
 import pathlib
-import re
 import stat
 
 import lsst.daf.butler as dafButler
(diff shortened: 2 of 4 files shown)


dhirving:
This passed jenkins but I'm not set up to run `ap_verify.py` myself -- would you mind running it for me?

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -277,7 +277,10 @@ def _getCollectionArguments(workspace, reuse):
     # skip-existing-in would work around that, but would lead to a worse bug in
     # the case that the user is alternating runs with and without --clean-run.
     # registry.refresh()
-    oldRuns = list(registry.queryCollections(re.compile(workspace.outputName + r"/\d+T\d+Z")))
+    collectionPattern = re.compile(workspace.outputName + r"/\d+T\d+Z")
+    oldRuns = list(registry.queryCollections(workspace.outputName + "/*"))
+    oldRuns = [run for run in oldRuns if collectionPattern.fullmatch(run)]
```
kfindeisen, line 282: Just to be sure I understand this -- regexes are being phased out, but globs are not?

python/lsst/ap/verify/workspace.py
```diff
@@ -219,10 +218,31 @@ def _ensureCollection(self, registry, name, collectionType):
             The type of collection to add. This field is ignored when
             testing if a collection exists.
         """
-        matchingCollections = list(registry.queryCollections(re.compile(name)))
-        if not matchingCollections:
+        if not self._doesCollectionExist(registry, name):
             registry.registerCollection(name, type=collectionType)
```
kfindeisen, line 222: Given that `registerCollection` is now(?) idempotent, I wonder if this code could be removed entirely...

python/lsst/ap/verify/pipeline_driver.py
```diff
@@ -277,7 +277,10 @@ def _getCollectionArguments(workspace, reuse):
     # skip-existing-in would work around that, but would lead to a worse bug in
     # the case that the user is alternating runs with and without --clean-run.
     # registry.refresh()
-    oldRuns = list(registry.queryCollections(re.compile(workspace.outputName + r"/\d+T\d+Z")))
+    collectionPattern = re.compile(workspace.outputName + r"/\d+T\d+Z")
+    oldRuns = list(registry.queryCollections(workspace.outputName + "/*"))
```
timj, line 281: Minor comment but does it make things faster or slower if we use a more constrained glob such as `/????????T*Z` or `/*T*Z`?

kfindeisen (approved):
Looks good, just one policy question.

python/lsst/ap/verify/pipeline_driver.py
kfindeisen: This is all happening in a small, temporary repo, so I don't think performance is an issue.

kfindeisen:
I've run `ap_verify` on `ap_verify_ci_dc2`, no problems. I ran it twice on the same workspace to ensure this code actually gets exercised.

For future reference, `stack-os-matrix` [now has a `ci_ap` target](https://developer.lsst.io/stack/jenkins-stack-os-matrix.html#running-a-stack-os-matrix-job), though I'm not sure exactly which pipelines it runs.

python/lsst/ap/verify/pipeline_driver.py
dhirving: Correct -- globs will continue to work.

python/lsst/ap/verify/workspace.py
dhirving: I think this one can.  The one below is less clear since it avoids modifying the chain if the collection already existed.

I'm gonna leave it alone to avoid needing to re-test, though.

PR #240, opened 2024-10-30
DM-47299: Add the afterburner subset to all of the pipelines

isullivan:

****

- [x] Do unit tests pass (`scons` and/or `stack-os-matrix`)?
- [x] Did you run `ap_verify.py` on at least one of [the standard datasets](https://pipelines.lsst.io/v/daily/modules/lsst.ap.verify/datasets.html#supported-datasets)?
- [x] Is the Sphinx documentation up-to-date?

### DIFF
--- a/pipelines/DECam/ApVerify.yaml
+++ b/pipelines/DECam/ApVerify.yaml
@@ -8,9 +8,11 @@ imports:
     # a metrics subset because it would require constant micromanagement.
     exclude:
       - prompt
+      - afterburner
   - location: $AP_PIPE_DIR/pipelines/DECam/ApPipe.yaml
     include:
       - prompt
+      - afterburner
 tasks:
   # ApVerify override removed by excluding apPipe.
   diaPipe:
--- a/pipelines/DECam/ApVerifyWithFakes.yaml
+++ b/pipelines/DECam/ApVerifyWithFakes.yaml
@@ -12,9 +12,11 @@ imports:
     # a metrics subset because it would require constant micromanagement.
     exclude:
       - prompt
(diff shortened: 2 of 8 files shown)


kfindeisen (commented):
Strictly speaking, the instrument-specific `ApVerify` should splice in the `afterburner` subset from the instrument-specific `ApPipe` -- as it is, the `ApVerify` will drop any overrides to afterburner tasks from `ap_pipe`. I'd recommend doing that for consistency's sake, even though the odds of such overrides existing are pretty low.

The problem would be if we end up overriding any of those tasks _for ap_verify_, like we do with `diaPipe`...

isullivan:
Good point. I have moved the import to the instrument-specific `ApVerify` definitions.

kfindeisen:
I hope you meant copy (`_instrument` should still be a representative pipeline), and the same goes for `ApVerifyWithFakes`.

pipelines/DECam/ApVerify.yaml
```diff
@@ -11,6 +11,7 @@ imports:
   - location: $AP_PIPE_DIR/pipelines/DECam/ApPipe.yaml
     include:
       - prompt
+      - afterburner
```
kfindeisen, line 14: You'd need to also exclude from `_ingredients/ApVerify.yaml`, else you have double definitions.

isullivan:
I had moved the imports rather than copied and excluded them, but I have fixed that up now.

kfindeisen (approved):
Looks good, but please confirm that the final pipelines are still valid.

PR #245, opened 2025-02-19
DM-46249: Fix test code to take into account new sasquatch API changes

timj:
{Summary of changes. Prefix PR title with JIRA issue.}

****

- [x] Do unit tests pass (`scons` and/or `stack-os-matrix`)?
- [ ] Did you run `ap_verify.py` on at least one of [the standard datasets](https://pipelines.lsst.io/v/daily/modules/lsst.ap.verify/datasets.html#supported-datasets)?
      For changes to metrics, the `print_metricvalues` script from `lsst.verify` will be useful.
- [ ] Is the Sphinx documentation up-to-date?

PR #244, opened 2025-02-17
DM-46249: Add SasquatchDatastore to enable uploading analysis_tools metrics

BrunoSanchez:
Update ci run script to take namespace and url for sasquatch upload."

Add optional ChainedDatastore to host SasquatchDatastore

Add new parameters to accomodate for Sasquatch metric upload

{Summary of changes. Prefix PR title with JIRA issue.}

****

- [x] Do unit tests pass (`scons` and/or `stack-os-matrix`)?
- [x] Did you run `ap_verify.py` on at least one of [the standard datasets](https://pipelines.lsst.io/v/daily/modules/lsst.ap.verify/datasets.html#supported-datasets)?
      For changes to metrics, the `print_metricvalues` script from `lsst.verify` will be useful.
- [x] Is the Sphinx documentation up-to-date?

PR #246, opened 2025-02-19
DM-46249: Hotfix for Sasquatch parameters and unit tests

kfindeisen:
This PR reverts the test changes on #245 (as we should be able to run `ap_verify` without access to Sasquatch) and instead makes the relevant parameters optional. It also adds documentation missing from #244.

****


BrunoSanchez (approved):
Looks good to me

PR #249, opened 2025-04-15
DM-49875: rename task labels and dataset names for RFC-1088

isullivan:
Renaming the datasets without updating the task defaults wreaks havoc on `ap_verify`, where we have to redefine diaPipe (now associateApdb) in many places. We should create a common config for associateApdb to simplify these definitions!

****

- [ ] Do unit tests pass (`scons` and/or `stack-os-matrix`)?
- [ ] Did you run `ap_verify.py` on at least one of [the standard datasets](https://pipelines.lsst.io/v/daily/modules/lsst.ap.verify/datasets.html#supported-datasets)?

### DIFF
--- a/doc/lsst.ap.verify/command-line-reference.rst
+++ b/doc/lsst.ap.verify/command-line-reference.rst
@@ -129,7 +129,7 @@ Required arguments are :option:`--dataset` and :option:`--output`.
 
    .. note::
 
-      At present, ap_verify assumes that the provided pipeline includes the ``diaPipe`` task from the AP pipeline, and configures it on the fly.
+      At present, ap_verify assumes that the provided pipeline includes the ``apdb_config`` parameter, which should be a path to the file created by ``apdb-cli create-cassandra`` or ``apdb-cli create-sql`` when initializing the APDB.
       It will likely crash if this task is missing.
 
 .. option:: --extra <key=value>
--- a/pipelines/DECam/ApVerify.yaml
+++ b/pipelines/DECam/ApVerify.yaml
@@ -15,11 +15,11 @@ imports:
       - afterburner
 tasks:
   # ApVerify override removed by excluding apPipe.
-  diaPipe:
+  associateApdb:
     class: lsst.ap.association.DiaPipelineTask
     config:
       doPackageAlerts: True
       alertPackager.doWriteAlerts: True
 contracts:
   # Contracts removed by excluding apPipe
-  - diaPipe.apdb_config_url == totalUnassociatedDiaObjects.apdb_config_url
+  - associateApdb.apdb_config_url == totalUnassociatedDiaObjects.apdb_config_url
--- a/pipelines/_ingredients/Conversions.yaml
+++ b/pipelines/_ingredients/Conversions.yaml
@@ -5,37 +5,37 @@
 description: Type conversion tasks customized for AP pipeline, with CalibrateImage.
 
 tasks:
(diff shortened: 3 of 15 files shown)


pipelines/DECam/ApVerify.yaml
```diff
@@ -15,11 +15,28 @@ imports:
       - afterburner
 tasks:
   # ApVerify override removed by excluding apPipe.
-  diaPipe:
+  associateApdb:
     class: lsst.ap.association.DiaPipelineTask
     config:
+      connections.coaddName: parameters.coaddName
```
TallJimbo, line 21: Shouldn't this connection override (and many others like it, in this and other pipelines) already be inherited from `_ingredients/ApVerify.yaml`?

TallJimbo (approved):

pipelines/DECam/ApVerify.yaml
kfindeisen: See comment on line 17: the process of splicing in $AP_PIPE_DIR/pipelines/DECam/ApPipe.yaml throws out those changes.

However, I would ask the same question in general -- shouldn't most of these overrides be inherited from (some flavor of) `ApPipe.yaml`?

doc/lsst.ap.verify/command-line-reference.rst
```diff
@@ -129,7 +129,7 @@ Required arguments are :option:`--dataset` and :option:`--output`.
 
    .. note::
 
-      At present, ap_verify assumes that the provided pipeline includes the ``diaPipe`` task from the AP pipeline, and configures it on the fly.
+      At present, ap_verify assumes that the provided pipeline includes the ``associateApdb`` task from the AP pipeline, and configures it on the fly.
       It will likely crash if this task is missing.
```
kfindeisen, line 133: This paragraph is out of date. [The current code](https://github.com/lsst/ap_verify/blob/main/python/lsst/ap/verify/pipeline_driver.py#L241-L242) assumes an `apdb_config` pipeline parameter, which is a much more robust requirement (but should still be noted).

pipelines/_ingredients/Conversions.yaml
```diff
@@ -26,12 +26,16 @@ tasks:
     class: lsst.pipe.tasks.postprocess.ConsolidateVisitSummaryTask
     config:
       connections.calexp: initial_pvi
+      connections.visitSummary: visitSummary
+      connections.visitSummarySchema: visitSummary_schema
```
kfindeisen: I don't understand why this change is needed, and there's no explanation in the commit message.

kfindeisen (changes requested):
You asked me to find a cleaner implementation of the `associateApdb` connections overrides, but I don't understand why they're needed at all. Yes, the per-instrument pipelines need to copy any config overrides from `_ingredients/ApVerify`, but AFAICT _those_ aren't needed either. The ones in `ap_pipe` should be sufficient.

Likewise, the overrides in `ap_pipe` seem to have some internal redundancy, going by the quick look I took -- there's no splicing going on there, so they should only need to be overridden in `_ingredients` or for genuinely new tasks.

pipelines/_ingredients/Conversions.yaml
isullivan: I am defining all of the connections with their default names, before changing them to the new names in a later commit.

kfindeisen (approved):
Looks good! Just make sure you're not duplicating things in other packages -- `ap_pipe` is supposed to be the source of truth for the pipeline definition.

